PR #96, opened 2026-05-24
fix(compose): unblock default install — bridge vars use ${VAR:-} not ${VAR:?} (DE-305, #92)

### LINKED ISSUE #92: [Bug] `docker compose up` fails on a default install — profile-gated bridge services use `${VAR:?}` required-variable checks
### Prerequisites

- [x] I have searched existing issues for this bug.
- [x] This is not a security vulnerability (those go through the disclosure process in SECURITY.md).

### What happened?

A fresh install following the README Quick Start fails at `docker compose up`.

The `slack-bridge` and `teams-bridge` services are correctly gated behind Compose profiles (`profiles: ["slack"]` / `["teams"]`), but their `environment:` blocks use the `${VAR:?error}` required-variable form, e.g.:

```yaml
SLACK_CLIENT_ID: ${SLACK_CLIENT_ID:?SLACK_CLIENT_ID is required when slack profile is enabled}
```

Compose interpolates the entire compose file *before* applying profile filtering, so these `:?` checks fire even when the `slack`/`teams` profiles are inactive. A default `docker compose up -d` therefore aborts:

```
error while interpolating services.slack-bridge.environment.SLACK_CLIENT_ID: required variable SLACK_CLIENT_ID is missing a value
```

Resolving each variable just surfaces the next one in turn.

`grep -oE '\$\{[A-Z_]+:\?' docker-compose.yml | sort -u` returns 12 variables using the `:?` form. Four are legitimate core secrets (`POSTGRES_PASSWORD`, `MINIO_ROOT_PASSWORD`, `LQ_AI_GATEWAY_KEY`, `JWT_SECRET`). The other eight belong to the opt-in bridge services and should not block a default install: `SLACK_CLIENT_ID`, `SLACK_CLIENT_SECRET`, `SLACK_SIGNING_SECRET`, `LQ_AI_BRIDGE_TOKEN`, `LQ_AI_BRIDGE_PUBLIC_URL`, `MICROSOFT_APP_ID`, `MICROSOFT_APP_PASSWORD`, `LQ_AI_TEAMS_BRIDGE_PUBLIC_URL`.

### Expected behavior

The `.env.example` header states the four core variables are what "must be set before `docker compose up` will succeed," and the Slack/Teams sections say their variables can be left blank when those integrations are unused.

Following that, a plain `docker compose up -d` with only the four core variables set should bring up the seven core services without touching the opt-in bridges.

### Reproduction steps

1. `git clone --recurse-submodules https://github.com/LegalQuants/lq-ai.git`
2. `cd lq-ai && cp .env.example .env`
3. Set only the four documented required variables in `.env`: POSTGRES_PASSWORD, MINIO_ROOT_PASSWORD, LQ_AI_GATEWAY_KEY, JWT_SECRET. Leave the Slack/Teams variables blank as `.env.example` instructs.
4. Run `docker compose up -d` with no `--profile` flag.
5. Observe the interpolation error for SLACK_CLIENT_ID.

### LQ.AI version

v0.2.0 116012bfae9782cacc701c55ef1fec398985d23f

### Deployment mode

Mode 1 (cloud LLM keys)

### LLM provider (if applicable)

_No response_

### Environment

- OS: macOS Tahoe 26.4
- Docker version 29.4.3, build 055a478
- Docker Compose version v5.1.4
- Bug occurs at `docker compose` interpolation time, before any container starts

### Relevant logs

```shell
There are no container logs — the failure happens during compose-file interpolation, before any service starts. The complete error output:

error while interpolating services.slack-bridge.environment.SLACK_CLIENT_ID: required variable SLACK_CLIENT_ID is missing a value: SLACK_CLIENT_ID is required when slack profile is enabled
```

### Additional context

Suggested fix: change the eight bridge variables from `${VAR:?...}` to the default form `${VAR:-}` in `docker-compose.yml`, and move the presence check into each bridge container's own entrypoint/startup script — which only runs when the profile is active. This preserves the "required when the profile is enabled" guarantee without breaking the default install path.

Alternatives: ship the eight bridge variables with placeholder values in `.env.example`, or document that they need non-empty placeholders even when unused. The `:-` change is cleaner since it requires no action from operators who don't use the bridges.

Workaround for affected users until fixed: set the eight bridge variables to any non-empty placeholder value in `.env`. The bridge services stay off because their profiles remain inactive.

Kevin-Tucuxi:
Fixes #92 (tracked internally as **DE-305**).

## Problem
Docker Compose interpolates **every** service definition before applying `--profile` filtering, so the profile-gated `slack-bridge`/`teams-bridge` env vars declared with the required-error `${VAR:?msg}` form aborted a plain `docker compose up` — even for operators who never enable those profiles:

```
error while interpolating services.slack-bridge.environment.SLACK_CLIENT_ID: required variable SLACK_CLIENT_ID is missing a value
```

This contradicted `.env.example` (the bridge vars are documented as leave-blank-if-unused) and broke the README Quick Start for every fresh install. Reported independently — with an excellent, precise diagnosis — by @legalinnovationaggregator-coder.

## Fix
- **`docker-compose.yml`:** the 8 opt-in bridge vars change `${VAR:?…}` → `${VAR:-}`. The 4 core secrets (`POSTGRES_PASSWORD`, `MINIO_ROOT_PASSWORD`, `LQ_AI_GATEWAY_KEY`, `JWT_SECRET`) keep `${VAR:?}`.
- **The "required when the profile is active" guarantee moves into each bridge's `Settings`** via a `field_validator` that rejects empty/whitespace credentials. It runs only when the bridge container starts (profile enabled) and preserves the exact `"X is required when the <slack|teams> profile is enabled"` message — fail-fast at startup beats a bridge that boots broken and only errors at OAuth time. (This is precisely the reporter's suggested approach.)
- **Regression tests:** `slack-bridge/tests/test_config.py` (7), `teams-bridge/tests/test_config.py` (6).

## Verification
- `docker compose --env-file <only 4 core vars> config` → **succeeds** with no `--profile` (the exact reported scenario).
- `docker compose --profile slack config` → still resolves `slack-bridge` (empty default, caught by the startup validator).
- Both new test suites pass; `ruff format`/`ruff check` clean.

> Touches `docker-compose.yml` + the bridge services. The bridge suites aren't in the main CI matrix (like the repo-root `tests/`), so the new config tests run locally / on demand.

Closes #92

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -430,13 +430,13 @@ services:
         condition: service_healthy
     environment:
       # Slack-side credentials (operator obtains from Slack App admin UI)
-      SLACK_CLIENT_ID: ${SLACK_CLIENT_ID:?SLACK_CLIENT_ID is required when slack profile is enabled}
-      SLACK_CLIENT_SECRET: ${SLACK_CLIENT_SECRET:?SLACK_CLIENT_SECRET is required when slack profile is enabled}
-      SLACK_SIGNING_SECRET: ${SLACK_SIGNING_SECRET:?SLACK_SIGNING_SECRET is required when slack profile is enabled}
+      SLACK_CLIENT_ID: ${SLACK_CLIENT_ID:-}
+      SLACK_CLIENT_SECRET: ${SLACK_CLIENT_SECRET:-}
+      SLACK_SIGNING_SECRET: ${SLACK_SIGNING_SECRET:-}
       # LQ.AI-side coordinates
       LQ_AI_BACKEND_URL: ${LQ_AI_BACKEND_URL:-http://api:8000}
-      LQ_AI_BRIDGE_TOKEN: ${LQ_AI_BRIDGE_TOKEN:?LQ_AI_BRIDGE_TOKEN is required when slack profile is enabled}
-      LQ_AI_BRIDGE_PUBLIC_URL: ${LQ_AI_BRIDGE_PUBLIC_URL:?LQ_AI_BRIDGE_PUBLIC_URL is required when slack profile is enabled}
+      LQ_AI_BRIDGE_TOKEN: ${LQ_AI_BRIDGE_TOKEN:-}
+      LQ_AI_BRIDGE_PUBLIC_URL: ${LQ_AI_BRIDGE_PUBLIC_URL:-}
       # Observability (opt-in)
       OTEL_EXPORTER_OTLP_ENDPOINT: ${OTEL_EXPORTER_OTLP_ENDPOINT:-}
       OTEL_SERVICE_NAME: ${OTEL_SERVICE_NAME:-lq-ai-slack-bridge}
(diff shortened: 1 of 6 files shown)


legalinnovationaggregator-coder:
Thanks, Kevin.

Yvonne

PR #199, opened 2026-06-21
docs(prd): file DE-353 — web/OpenWebUI first-boot HF model fetch blocks the user-facing proxy

Kevin-Tucuxi:
Files **DE-353** from the v0.5.0 release-readiness verification. On a fresh deploy the `web` (OpenWebUI) container downloads ~30 HF files (its default local RAG embedding + Whisper models) before binding `:8080`, so the user-facing `proxy` is gated ~10 min until the fetch completes (self-heals; not a hard failure). LQ.AI uses its own embeddings, so the fetch is unnecessary — fix is to disable/redirect OpenWebUI's local model engines or bake the models. Distinct from DE-351 (ingest-worker docling). Doc-only.


Kevin-Tucuxi:
Superseded: DE-353 is being **fixed** (not deferred) in the launcher-first-run PR — `RAG_EMBEDDING_ENGINE` compose env stops the web boot model-fetch. Re-filing DE-353 as resolved + adding DE-354 (launcher reset-before-start) there.

PR #263, opened 2026-07-02
fix(web): attach chat files to the model and keep their status live

dropthejase:
## What this PR does

Fixes two bugs in the per-message attach-file flow: attached files never reached the model (the request body never included `file_ids`), and the attached-file status badge stayed frozen at "pending" forever after upload.

## Why

Uploading a file to a chat and asking the model about it silently failed — the model would answer as if no file was attached, with no error surfaced anywhere. Separately, the status badge never reflected real ingestion progress, so a user had no way to tell when a file was actually ready to be referenced.

## What this PR does *not* do

Does not touch the project-level attached-files feature (`attached_file_ids` on `Project`, surfaced read-only via `projectFiles`) — that's a separate, already-working code path. Does not add a UI affordance for manually retrying a failed ingestion; that's out of scope.

## Type of change

## Testing

`web/src/lib/lq-ai/__tests__/ChatPanel-attach-file.test.ts` — 10 tests covering `hasPendingFileStatus()` and `fileIdsForSend()`, the two pure functions extracted into `ChatPanel.svelte`'s module block (mirroring the existing `detectSlashAt` convention, since `@testing-library/svelte` is unavailable for this file — see that test's header for the rationale).

<details>
<summary>Manual testing notes</summary>

Verified end-to-end against a locally running stack:
1. Uploaded a PDF to a chat, watched the attached-files panel badge transition from "pending" to "ready" within a few seconds (previously stayed on "pending" indefinitely).
2. Asked the model a question about the document's content — it answered from the real document text (previously it had no access to the file at all).
3. Confirmed via the gateway's audit log that `inference.message_files_attached` events show `injected_count` matching `attached_count` for the turn.
4. Confirmed via direct DB inspection that `documents.normalized_content` was populated and the ~33K-character contract text was genuinely being injected as a system message before the user turn.

</details>

## Documentation

*(no boxes checked — internal bugfix, no user-facing capability or doc change)*

- [x] **Egress (P1):** n/a — no new outbound call; uses the existing `GET /files/{id}` endpoint already called elsewhere in this file.
- [x] **Closed set (P2):** n/a — no new model/autonomous capability.
- [x] **No raw payloads (P3):** n/a — no new logging.
- [x] **Fail restrictive (P4):** the status-poll `catch` block logs and leaves the badge at its last-known state rather than throwing; a failed poll doesn't break the chat.
- [x] **Atomic audit (P5):** n/a — no new state-changing endpoint.
- [x] **One governance path (P6):** n/a.
- [x] **Human gate (P7):** n/a — no destructive action.
- [x] **Operator control (P8):** n/a.
- [x] **User owns data (P9):** n/a — no new data flow; `file_ids` uses the backend's pre-existing, already-anonymization-aware injection path.
- [x] **Contract is truth (P10):** n/a — `file_ids` was already documented in the backend's `MessageCreateRequest` schema; this PR only wires the frontend to use it.

## DCO sign-off

- [x] All commits in this PR are signed off per the DCO

## Related issues

None filed — found and fixed during manual testing of an unrelated feature.

## Reviewer notes

Two independent root causes in the same file, landed together since they were found and fixed in one pass:

1. `sendMessage()` never included `file_ids: chatFiles.map(f => f.id)` in the `messagesApi.sendMessageStream()` call — the backend's `file_ids` channel (`api/app/schemas/chats.py::MessageCreateRequest.file_ids`) was fully implemented and unused.
2. `uploadAttached()` pushed the upload-response file object into `chatFiles` once and nothing ever re-fetched it, so `ingestion_status` stayed frozen at its upload-time value. Added lightweight polling (3s interval, stops once every attached file reaches a terminal status) rather than reusing the existing chat-streaming SSE infrastructure — smaller diff, matches this repo's other non-realtime status patterns.

Would appreciate a second look at the polling interval/approach if there's a preference for SSE instead.

### DIFF
--- a/web/src/lib/lq-ai/__tests__/ChatPanel-attach-file.test.ts
+++ b/web/src/lib/lq-ai/__tests__/ChatPanel-attach-file.test.ts
@@ -1,0 +1,85 @@
+/**
+ * Regression tests for the chat-attach-file-context bugfix.
+ *
+ * Convention note: matches the ChatPanel-slash-detect.test.ts pattern —
+ * pure helpers are exported from <script context="module"> in the .svelte
+ * file and exercised here without mounting the (large) component, since
+ * @testing-library/svelte is unavailable for this file (see that test's
+ * header for the full rationale).
+ *
+ * Covers two fixes:
+ *   - Attached chat files never reached the model because sendMessage()
+ *     never included file_ids in the request body. fileIdsForSend() is
+ *     the extracted mapping.
+ *   - The attached-file status badge stayed frozen at "pending" forever
+ *     because nothing re-fetched ingestion_status after upload.
+ *     hasPendingFileStatus() is the extracted predicate the polling loop
+ *     keys off.
+ */
+import { describe, expect, it } from 'vitest';
+
+import { hasPendingFileStatus, fileIdsForSend } from '../components/ChatPanel.svelte';
+import type { FileMeta } from '../types';
+
+function makeFile(overrides: Partial<FileMeta> = {}): FileMeta {
+	return {
+		id: 'file-1',
+		owner_id: 'user-1',
+		filename: 'contract.pdf',
+		mime_type: 'application/pdf',
+		size_bytes: 1024,
+		ingestion_status: 'pending',
+		created_at: '2026-01-01T00:00:00Z',
+		...overrides
+	};
+}
+
+describe('hasPendingFileStatus', () => {
+	it('returns false for an empty file list', () => {
+		expect(hasPendingFileStatus([])).toBe(false);
+	});
+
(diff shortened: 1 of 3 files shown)


houfu:
Hi @dropthejase — I'm going to close this in favour of #402, which fixes the same two things and carries them further: the helpers live in their own tested module, the server's 16-file cap is enforced client-side, and the attached-file list is cleared on chat switch.

You found a real silent failure here — a backend channel the frontend never called — and traced it through the audit log rather than stopping at the symptom. You were also first to it, and then waited six weeks for a reply. That's on us, and I'm sorry.

Nothing needed from you. #264 is still open and still yours; that's the one I'll pick up next.

PR #264, opened 2026-07-02
fix(web): persist per-chat model selection across page refresh

dropthejase:
## What this PR does

Persists a chat's selected model across a page refresh, by saving it to `localStorage`.

## Why

If a user picked a model for a chat and then refreshed the page, their choice was silently thrown away — the chat reverted to the default model with no warning. Root cause: the selected model was only ever held in memory (`modelByChat`, a component-local `Record`), never saved anywhere durable. A refresh remounts the component from scratch, so that in-memory value resets, and the picker's fallback logic quietly picks the default instead.

## What this PR does *not* do

Does not change the per-chat scoping of model selection (a deliberate, pre-existing design — different chats can already use different models; this PR only makes each chat's choice survive a refresh). Does not add cross-device/cross-browser persistence — this is a `localStorage` fix, scoped to one browser on one device, matching the severity of the bug.

## Type of change

## Testing

`web/src/lib/lq-ai/__tests__/ChatPanel-model-persist.test.ts` — 7 tests, mirroring `ReceiptsDrawer.test.ts`'s `MockStorage` pattern exactly (same file already does per-chat `localStorage` persistence for a different feature — the receipts-drawer open state). Covers key-namespacing, read/write, cross-chat isolation, overwrite, and round-trip.

<details>
<summary>Manual testing notes</summary>

Verified against a locally running stack: picked a non-default model for a chat, sent a message, hard-refreshed the page — the picker still showed the previously-selected model instead of resetting to the default (`anthropic-prod`/`claude-opus-4-7`).

Also verified the race-condition concern doesn't apply: `loadShell()` (which populates `availableModels`, the picker's source list) is `await`ed in `onMount` before `selectChat` ever runs, so by the time a persisted model id is hydrated from `localStorage`, the picker's model list is already loaded.

</details>

## Documentation

*(no boxes checked — internal bugfix, no user-facing capability or doc change)*

- [x] **Egress (P1):** n/a — no outbound call involved; `localStorage` is a browser-local read/write.
- [x] **Closed set (P2):** n/a.
- [x] **No raw payloads (P3):** n/a — only a model-id string is persisted, no message content.
- [x] **Fail restrictive (P4):** `readPersistedModel`/`writePersistedModel` both no-op safely if `localStorage` is unavailable (SSR/private-browsing edge cases) — falls through to the existing default-selection behavior.
- [x] **Atomic audit (P5):** n/a.
- [x] **One governance path (P6):** n/a.
- [x] **Human gate (P7):** n/a.
- [x] **Operator control (P8):** n/a.
- [x] **User owns data (P9):** n/a — client-local preference only, not sent to the backend.
- [x] **Contract is truth (P10):** n/a — no API/schema change.

## DCO sign-off

- [x] All commits in this PR are signed off per the DCO

## Related issues

None filed — found during manual testing of an unrelated feature.

## Reviewer notes

`selectModel()` write-throughs to `localStorage` on every pick; `selectChat()` hydrates `modelByChat[chat.id]` from `localStorage` only when no in-session choice already exists, so a same-session pick is never clobbered by a stale persisted value.

### DIFF
--- a/web/src/lib/lq-ai/__tests__/ChatPanel-model-persist.test.ts
+++ b/web/src/lib/lq-ai/__tests__/ChatPanel-model-persist.test.ts
@@ -1,0 +1,96 @@
+/**
+ * Regression test for the model-selection-persist-refresh bugfix.
+ *
+ * Convention note: mirrors ReceiptsDrawer.test.ts exactly — pure
+ * localStorage helpers exported from <script context="module"> in the
+ * .svelte file, exercised here with an injectable MockStorage, without
+ * mounting the (large) component (@testing-library/svelte is unavailable
+ * for this file — see ChatPanel-slash-detect.test.ts header).
+ *
+ * Bug: modelByChat was a component-local Record with no persistence, so a
+ * full page refresh remounted the component, reset modelByChat to {}, and
+ * currentModelId's reactive fallback silently picked the picker's default
+ * — discarding the user's chosen model for that chat.
+ */
+import { describe, it, expect } from 'vitest';
+import {
+	modelStorageKeyForChat,
+	readPersistedModel,
+	writePersistedModel
+} from '../components/ChatPanel.svelte';
+
+class MockStorage implements Storage {
+	private map = new Map<string, string>();
+	get length() {
+		return this.map.size;
+	}
+	key(i: number): string | null {
+		return Array.from(this.map.keys())[i] ?? null;
+	}
+	getItem(k: string): string | null {
+		return this.map.get(k) ?? null;
+	}
+	setItem(k: string, v: string): void {
+		this.map.set(k, v);
+	}
+	removeItem(k: string): void {
+		this.map.delete(k);
+	}
+	clear(): void {
+		this.map.clear();
+	}
+}
+
+describe('modelStorageKeyForChat', () => {
+	it('namespaces by chat ID', () => {
+		expect(modelStorageKeyForChat('c1')).toBe('lq_ai_chat_model_c1');
+		expect(modelStorageKeyForChat('uuid-here')).toBe('lq_ai_chat_model_uuid-here');
+	});
+});
+
+describe('readPersistedModel', () => {
+	it('returns null for an unset chat', () => {
+		const s = new MockStorage();
+		expect(readPersistedModel('c1', s)).toBeNull();
+	});
+
+	it('returns the persisted model id', () => {
+		const s = new MockStorage();
+		s.setItem(modelStorageKeyForChat('c1'), 'anthropic-prod/claude-opus-4-8');
+		expect(readPersistedModel('c1', s)).toBe('anthropic-prod/claude-opus-4-8');
+	});
+
+	it('scopes reads per chat id (no cross-chat bleed)', () => {
+		const s = new MockStorage();
+		s.setItem(modelStorageKeyForChat('c1'), 'model-a');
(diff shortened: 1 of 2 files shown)


houfu (changes requested):
The chat-scoped persistence is a useful fix and the happy path is covered cleanly. Could you make the following focused update? Please wrap `readPersistedModel()` and `writePersistedModel()` in `try/catch`, returning `null`/no-op if Storage throws. Then extract a pure `readAvailablePersistedModel(chatId, availableIds, storage?)` helper that accepts the saved ID only when it appears in the supplied catalog, and use it from `selectChat()` with the IDs in `availableModels.data`. Four small tests would pin the boundary: throwing `getItem`, throwing `setItem`, a catalog hit, and a stale catalog miss. This keeps chat selection usable when browser storage fails and ensures the picker displays the same model that the send path uses.

houfu:
Hi @dropthejase — following up on this one. The chat-scoped persistence fix is solid and I'd like to get it into v0.7.1. The outstanding ask is still what I left on Aug 13: wrap `readPersistedModel()`/`writePersistedModel()` in try/catch so a `localStorage` failure doesn't break chat selection, and extract a pure `readAvailablePersistedModel(chatId, availableIds, storage?)` helper that only accepts a saved model id if it's in the current catalog — plus four small tests (throwing `getItem`, throwing `setItem`, a catalog hit, a stale catalog miss).

I'm cutting the v0.7.1 release this weekend. If you can push an update by **Sunday**, I'll get this merged as part of it. If you don't have the bandwidth right now, that's completely fine — I can add the try/catch and helper directly on your branch (keeping you as the author), or I'll move this to v0.7.2 so it's ready whenever you're back to it.

Either way, thanks for catching and fixing this — it's a real bug and a clean root-cause fix.

dropthejase:
Thanks for the follow-up. Pushed the requested changes and rebased onto current `main` (was `d0860b00`, now `5ff95667`).

**The two changes you asked for:**
- `readPersistedModel` / `writePersistedModel` now wrap the `getItem` / `setItem` call in `try/catch` — a `Storage` that throws (Safari private mode, disabled-storage policy, quota-exceeded write) returns `null` / no-ops instead of propagating and breaking chat selection. Falls through to the existing default-selection behaviour.
- New pure helper `readAvailablePersistedModel(chatId, availableIds, storage?)` returns the saved id only when it is still in the supplied catalog. `selectChat()` now calls it with the ids from `availableModels.data`, so a model the operator has since removed from the gateway is never surfaced by the picker while the send path silently falls back to the default alias — picker and request stay in lockstep.

**Four tests** pin the boundary: throwing `getItem`, throwing `setItem`, a catalog hit, and a stale catalog miss.

**Rebase note:** `main` commit `e3625cce` ("send chat-attached file_ids and poll ingestion status") had added its own block to `selectChat()` at the same spot. Resolved by keeping both — its `abortFilePolls(); chatFiles = []` reset runs first, then the model-hydration block. They are independent (attached-file draft state vs. `modelByChat`), so ordering does not matter functionally.

Local verification: full web suite green (768 tests), `svelte-check` 0 errors, Prettier clean on the changed files. The 4 pre-existing `_interactionId` / `_msg` ESLint errors in `ChatPanel.svelte` are untouched.

houfu (approved):
Thanks for the thorough follow-up—this addresses the storage-failure and stale-model cases I called out, and the focused tests cover each boundary. Approved from my review;  this is ready to merge.

PR #282, opened 2026-07-06
docs(adr): propose ADR 0036 — lq design-system primitives for the /lq-ai shell

sgbooth:
## Summary

- Adds a small primitive layer (`LqButton`, `LqGroup`, `LqStack`, `LqText`, `LqTitle`, `LqAlert`, `LqPill`, `LqSwitch`, `LqCard`, `LqHoverCard`) under `web/src/lib/lq-ai/components/shared/`, backed by one token file (`shared/types.ts`: variant/tone/size/gap/shadow/weight/justify/align scales, each a `const` array with a derived type for full editor autocomplete).
- Modeled on Mantine's component API shape (`variant`/`size`/`tone`, `Component.Part` compound namespacing) as a naming/ergonomics reference only — no framework dependency added. `LqHoverCard` reuses `bits-ui`'s headless `Portal`/`Switch` (already a dependency) rather than reimplementing positioning/accessibility behavior.
- Migrates `ChatPanel.svelte`, `MessageBubble.svelte`, and `MessageList.svelte` off ad hoc Tailwind spacing/color classes onto these primitives.
- Converts two **display-only** modals (`EnhancedDiffModal`, `CitationLedgerPanel`) into hover-triggered `LqHoverCard` popovers, portaled to `document.body` so they aren't clipped by the message list's `overflow-y-auto` scroll container. Modals with real mutation/interaction (`AttachKBModal`, `TierFloorOverrideModal`, `CaptureSkillModal`) were evaluated and deliberately left as modals — hover-triggering an audited security override, for example, would be a real regression, not a style choice.
- Deduplicates a tier→description mapping that had drifted across three independent copies (`TierBadge`, `TierDetailsPanel`, `MessageBubble`) into one source of truth, `chat/tierDescriptions.ts`.
- Adds `@tabler/icons-svelte` as the icon source for these primitives (new dependency — see ADR 0022 for justification: emoji-as-icons force inline-style hacks per call site to size/align them, which is the exact inconsistent-spacing/sizing problem this effort targets).
- Updates `CLAUDE.md` / `CONTRIBUTING.md`'s "use the project's design system primitives" line (previously pointed at nothing concrete for `/lq-ai`) to name this layer, and adds **ADR 0022** documenting the decision, the governing rules for new primitives/tokens, and follow-up work (OKLCH color migration, `LqPopover`, `LqSimpleGrid`).

Scoped entirely to `web/src/lib/lq-ai/**` — the OpenWebUI shell at `/` is untouched per ADR 0009.

## Why

`lib/lq-ai/components/**` had no shared vocabulary for spacing, color, sizing, or interaction — every component decided these independently in raw Tailwind utility classes. That wasn't a problem in any single file, but it compounded:

- **Duplicated, drifting logic.** The tier→description mapping (label, blurb, tone per inference tier) was hand-copied into three separate files — `TierBadge.svelte`, `TierDetailsPanel.svelte`, and `MessageBubble.svelte` — with the wording already diverging between copies. Nothing enforced that a change to "what Tier 3 means" would land in all three places.
- **Inconsistent spacing from self-margins instead of layout gap.** Sibling elements used ad hoc `mt-1`/`mt-2`/`mb-3` classes to create breathing room, chosen per element by feel rather than from one spacing scale — so adjacent elements in the same list ended up with 4px, 8px, or 0px gaps depending on which developer touched which line, not on any actual design intent.
- **A real bug, not just an aesthetic one.** `ChatPanel.svelte` had a `gap={2}` typo on an ad hoc flex wrapper that silently fell through to an unintended raw-pixel value instead of erroring, because there was no typed vocabulary to catch it.
- **Emoji as icons forced inline-style hacks.** Several components used emoji glyphs (🛠️🔒🧠📎📜✨) as icons, which have no `size` prop and no predictable box model — every call site ended up hand-tuning `font-size`/`line-height`/margin to make them sit correctly next to text, which is the same "inconsistent spacing via ad hoc styling" problem as the margin issue above, just applied to icons instead of layout.
- **The existing convention pointed at nothing.** CONTRIBUTING.md and CLAUDE.md have both said "use the project's design system primitives rather than ad-hoc Tailwind" since M1, but nothing concrete existed for `/lq-ai/*` to point to — so the guidance was unenforceable in practice.

This PR builds that shared vocabulary (a token file plus a small set of Svelte primitives), fixes the concrete bugs found while migrating real call sites onto it, and updates the convention docs to point at something real instead of nothing. See ADR 0022 for the full design rationale and the rules for extending it.

## What this PR does *not* do

Does not rewrite all existing components using new components.  Only selected components are modified to illustrate improved semantics, improved hierarchy, and reduced interaction/logic. Does not implement a color system that has semantic consistency, this is to do in later updates.

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds capability)
- [ ] Documentation update
- [x] Skill contribution (see attestation below)
- [ ] Test / CI / tooling

## Testing

- [ ] Unit tests added/updated
- [x] Integration tests added/updated (if applicable)
- [x] Manually tested against a real deployment (describe scenario below)

<details>
<summary>Manual testing notes (if applicable)</summary>

Smoke tested using live data to validate correct layout.

</details>

## Documentation

- [x] Skill-authoring guide updated (if skill conventions change)

- [ ] **Egress (P1):** any outbound third-party call goes through the gateway only — no direct egress from `api/`.
- [ ] **Closed set (P2):** any new model/autonomous-invokable capability is a bounded, operator-controlled allowlist entry, not an open hook.
- [ ] **No raw payloads (P3):** no row or log line I add carries raw content (message bodies, tool args/results, fetched text, keys) — counts, types, ids, digests, outcomes only.
- [ ] **Fail restrictive (P4):** missing/broken config fails closed, and I tested that path, not just the happy path.
- [ ] **Atomic audit (P5):** state changes get an audit row committed in the same transaction (helpers flush, the handler commits).
- [ ] **One governance path (P6):** I reused the shared governance/audit helpers rather than re-deriving tier/cost/audit logic.
- [ ] **Human gate (P7):** anything destructive/irreversible is behind the confirmation gate and excluded from unattended/autonomous execution.
- [ ] **Operator control (P8):** any new capability has an operator enable/disable and appears on an admin surface.
- [ ] **User owns data (P9):** outbound matter context is anonymized; the change respects export/delete and data-residency guarantees.
- [ ] **Contract is truth (P10):** OpenAPI sketch / DB schema doc / relevant ADR updated in this PR; any architectural/authz decision is recorded or surfaced, not made silently.

## DCO sign-off

- [x] All commits in this PR are signed off per the [DCO](../CONTRIBUTING.md#sign-off-developer-certificate-of-origin)

## Related issues

---

## Skill update

Updates the skills to apply the semantics from the design primitives for consistency and improved semantics.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -82,7 +82,7 @@ Each subsystem (`api/`, `gateway/`, `web/`) is a self-contained service. They ta
 - **Linter:** ESLint per `.eslintrc`.
 - **TypeScript:** required for new files; legacy `.js` files migrate gradually.
 - **Framework:** SvelteKit. The `web/` subdirectory is a fork of OpenWebUI, which is a SvelteKit app — extensions and customizations stay in Svelte. We do **not** mix React into `web/`. The Word add-in (`word-addin/`, M3) uses Office.js with React; the `web/` codebase does not.
-- **Component conventions:** match the OpenWebUI conventions for shared components; use the project's design system primitives rather than ad-hoc Tailwind.
+- **Component conventions:** match the OpenWebUI conventions for shared components (`web/src/lib/components/**`, untouched per ADR 0009). For everything under `web/src/lib/lq-ai/**`, use the `lq` design-system primitives (`web/src/lib/lq-ai/components/shared/` — `LqButton`, `LqGroup`, `LqStack`, `LqText`, `LqTitle`, `LqAlert`, `LqPill`, `LqSwitch`, `LqCard`, `LqHoverCard`) and the shared token file (`shared/types.ts`) rather than ad-hoc Tailwind spacing/color/size classes. See [ADR 0022](docs/adr/0022-lq-design-system-primitives.md) for the full rationale and the rules for adding new primitives/tokens.
 
 ### Both
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -149,7 +149,7 @@ git rebase --signoff main                 # all commits since main diverged
 - **Linter:** ESLint with the project's `.eslintrc`. TypeScript-specific rules enforced for `.ts` files; Svelte rules for `.svelte` files.
 - **TypeScript:** required for new files; gradual migration of legacy `.js` files welcome but not required for unrelated changes.
 - **Framework:** SvelteKit. The `web/` codebase is a fork of OpenWebUI, which is a SvelteKit app — extensions and customizations stay in Svelte. Do **not** introduce React into `web/`. The Word add-in (`word-addin/`, M3) uses Office.js with React; the `web/` codebase does not.
-- **Component conventions:** match the OpenWebUI conventions for shared components; use the project's design system primitives rather than ad-hoc Tailwind.
+- **Component conventions:** match the OpenWebUI conventions for shared components (`web/src/lib/components/**`, untouched per ADR 0009). For everything under `web/src/lib/lq-ai/**`, use the `lq` design-system primitives (`web/src/lib/lq-ai/components/shared/` — `LqButton`, `LqGroup`, `LqStack`, `LqText`, `LqTitle`, `LqAlert`, `LqPill`, `LqSwitch`, `LqCard`, `LqHoverCard`) and the shared token file (`shared/types.ts`) rather than ad-hoc Tailwind spacing/color/size classes. See [ADR 0022](docs/adr/0022-lq-design-system-primitives.md) for the full rationale and the rules for adding new primitives/tokens.
 
 ### Configuration files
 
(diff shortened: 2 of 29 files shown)


houfu:
Thanks for your patience on this, @sgbooth — it's sat since July and that's on us, not on anything you did. LQ Members do ask about the look of LQ.AI, and I think you are going in the right direction by thinking about primitives, which will help in redesign the overall UI of the application. However, we're going to take our time on this. Here's what I propose to do (not everything is in here right now)

1. Redo this as an ADR. We will need to take some measured thinking of how to do design "primitives". With Claude Design and similar tools, this is less expensive than it used to be, but I still need to consult community members who are into web and UI design what their preferences are.
2. Port the ones you've made in accordance with the ADR as decided.
3. Begin to port the rest of the 90% of components which need to be ported as a separate punchless/checklist.

Thanks for contributing!

Issue #275, opened 2026-07-04
DE-358 — tool-use hardening backlog (PRD §9)


SaifAlYounan:
Claiming DE-358 per the roadmap process. This is the backlog Keller filed from PRs #185/#186 (tracking issue #215). Six items in PRD §9, all P3 and S effort.

Plan: fresh branch off main. Items 2 to 4 first (OpenAPI doc of tools/tool_choice, tools count cap, granular tool_choice tests). Then verify the streaming tool_use gap is reachable before fixing it. Fernet encryption of the resume payload and the api-side tier ceiling as the last two.

I'd like to take this.

SaifAlYounan:
Items 1–4 are up as two PRs: #284 (items 2–4: OpenAPI docs, tools cap at 64, per-mode tool_choice tests) and #285 (item 1: streaming tool_use accumulator — reachability verified first per the DE entry; the gap is latent, the chat loop is non-streaming and nothing combines stream+tools today).

Items 5 and 6 each hide a decision I'd rather surface than make unilaterally (P10), so putting both here before building.

**Item 5 — Fernet encryption of the pending-tool-call resume payload.** Current state is plaintext JSONB by design: the model docstring classifies `tool_call_args` / `resume_state` as the same sensitivity class as `messages.content`, which is also plaintext. Encrypting is roughly a day of work (the `MCPTokenEncryptor` Fernet helper and its key path already exist), but it buys a *higher* at-rest bar for the pending-call store than for message content itself, and it costs: the api process gains `LQ_AI_MCP_MASTER_KEY` as a hard requirement outside the MCP-OAuth feature, a migration (0066) with an encrypt-backfill for in-flight pending calls, and the resume path gains a decrypt failure mode (key rotation breaks pending confirmations). My read: **decline** unless we want the pending-call store held to a higher bar than messages — in which case the consistent move is encrypting both, which is a bigger conversation (operator-managed at-rest encryption is the documented posture today). If declined I'll PR the DE-358 entry update recording "considered, declined, reasoning" so the sub-item closes honestly.

[...]

PR #285, opened 2026-07-07
DE-358 item 1: accumulate streamed Anthropic tool_use into tool_calls deltas

SaifAlYounan:
Part of DE-358 (PRD §9 tool-use hardening backlog), refs #275. Rebuilt against current main.

**Reachability verified before fixing, per the DE entry:** the gap is latent, not an active bug. The chat tool-loop drives non-streaming gateway calls (`run_chat_tool_loop` sends `stream: False`), and no consumer in `api/`, `web/`, or `tests/` combines `stream=true` with `tools`. This PR is forward-hardening so the first streaming-with-tools consumer does not silently lose tool calls.

`_anthropic_stream_iter` previously translated only `text_delta` blocks; a streamed `tool_use` block produced no `tool_calls` delta and no error. Now: `content_block_start` (type `tool_use`) opens an accumulator keyed by block index, `input_json_delta` fragments append, and one OpenAI-shaped `tool_calls` delta is emitted at `content_block_stop` — mirroring the non-streaming bridge in `_from_anthropic_response` (arguments as a JSON string; empty input becomes `"{}"`). The existing `tool_use → tool_calls` finish-reason mapping now fires on a populated stream.

Adds the suite's first streaming-tool tests: multi-fragment accumulation and the no-argument tool call. Text-only streams are byte-identical (the existing streaming test is untouched and green).

**Contributor checklist:** no new egress; no payload logged; tests in this PR; no new dependency. `gateway/**` routes to security review per CODEOWNERS.

Gates run locally: `ruff format` + `ruff check` clean, `mypy --strict` gateway clean, adapter test file 26/26 green.

### DIFF
--- a/gateway/app/providers/anthropic.py
+++ b/gateway/app/providers/anthropic.py
@@ -575,6 +575,9 @@ async def _anthropic_stream_iter(
 
     * One initial chunk with ``delta.role = "assistant"``.
     * One chunk per text delta with ``delta.content = "<text piece>"``.
+    * One chunk per completed ``tool_use`` block with
+      ``delta.tool_calls`` (DE-358 item 1: accumulated from
+      ``input_json_delta`` fragments, emitted at ``content_block_stop``).
     * One final chunk with ``finish_reason`` set, ``delta`` empty, and
       a ``usage`` block.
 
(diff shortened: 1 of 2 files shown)


SaifAlYounan:
Superseded by #395 — this PR was auto-closed when I accidentally deleted my fork, and GitHub cannot
commit** (`7280661e67c5`); no content changed.

Issue #288, opened 2026-07-07
Security Audit (Claude Fable 5): 22 findings across gateway / api / web / supply-chain


SaifAlYounan:
# LQ.AI — Full 360° Security Audit (Claude Fable 5)

> **Note on disclosure:** LQ.AI's `SECURITY.md` asks for coordinated disclosure of vulnerabilities. This report is being posted as a public issue as a deliberate choice by an LQAI committee coordinator, because none of the findings below is exploitable by an external attacker in the shipped default configuration (every service binds to `127.0.0.1`), the findings are hardening gaps rather than live exploits, and public visibility serves the committee's review. The full technical summary has also been shared in the committee's private channel. Maintainers who would prefer this moved to a private advisory: say so and it will be.

> **Automated security audit by Claude Fable 5.** This report was produced by a multi-agent harness: five domain auditors swept the **gateway, api, web, desktop, bridges, skills, CI, and docker** surfaces of the self-hosted LQ.AI legal-AI platform, and every high/critical candidate was then subjected to per-finding **adversarial verification** against the platform's real deployment posture — a default docker-compose that binds all services to `127.0.0.1`. Verification either confirmed a finding in code, downgraded its severity to match true reachability, or refuted it. Severities below reflect that adjudication, not the raw first-pass ratings. **Scope:** application and infrastructure source in the LQ.AI monorepo (gateway, api, web/OpenWebUI fork, desktop/Electron, Slack/Teams bridges, word-addin, skills subsystem, CI workflows, Dockerfiles, docker-compose). Out of scope: live penetration testing, dependency CVE fuzzing beyond version review, and the upstream Open WebUI codebase except where LQ.AI code depends on its behavior.

---

## Executive summary

LQ.AI is a **notably security-conscious codebase**. The design shows real defensive intent: a hardened tool-egress guard that defeats DNS rebinding, encrypted-at-rest provider keys with a master key never written to disk, constant-time secret comparisons, fail-closed config loading, a graph-driven (not model-driven) phase-grant system for the autonomous agent, and a docker-compose that binds every port to loopback by default. Many controls are done correctly and consistently. The findings below are, in large part, **gaps in an otherwise coherent defensive model** rather than a codebase that ignores security.

**Counts by severity (22 findings):**

| Severity | Count |
|---|---|
| High | 3 |
| Medium | 10 |
| Low | 8 |
| Info | 1 |

[...]

PR #292, opened 2026-07-07
Refuse to start with the published dev JWT secret in production (API-04, Refs #288)

SaifAlYounan:
## What this PR does

Makes the API refuse to start when `JWT_SECRET` is still the published development default (`dev-jwt-secret-change-me`) and dev mode is off.

## Why

`Refs #288` — finding **API-04 (Medium)**. `jwt_secret` defaulted to the hard-coded literal `dev-jwt-secret-change-me` and signs **and verifies** every HS256 access/MFA token. `Settings` had no validator and there was no startup guard, so an operator who forgot to set `JWT_SECRET` (a silent, missed `.env` line) would ship with a **public** signing key — an attacker can then forge an access token for any user id, including an admin's. The docstring told operators to override it, but that is documentation, not enforcement.

## How

- Add `assert_production_secrets(settings)` in `app/config.py` and call it from the `app.main.lifespan` startup gate. It raises `RuntimeError` when `jwt_secret == DEV_JWT_SECRET and not lq_ai_dev_mode`.
- Deliberately a **lifespan gate, not a Settings `@model_validator`** — construction with the published defaults must stay valid (e.g. `test_config.test_defaults_apply_when_env_unset`, and any `Settings()` in tests). No test enters the full `lifespan()` (the ASGI in-process client doesn't trigger it), so this doesn't disturb the suite; it fires only when the process actually serves.
- Extract the default into the `DEV_JWT_SECRET` constant so the `Field` default and the guard share one source of truth.
- `LQ_AI_DEV_MODE=true` remains the explicit escape hatch for local dev (the existing "relax some safety checks for local development" flag).

## What this PR does *not* do

Doesn't broaden the check to other secrets (gateway key etc.) — scoped to the JWT default called out in the finding. Doesn't change token logic.

## Type of change

## Testing

<details>
<summary>Gates run locally (api/.venv)</summary>

- `pytest tests/test_config.py` → **7 passed** (incl. the 3 new: dev-default-off-dev-mode → raises; dev-default-in-dev-mode → ok; real secret → ok). These are pure unit tests (no DB), so they ran locally.
- `mypy app` → `Success: no issues found in 190 source files`.

</details>

- [x] **Fail restrictive (P4):** a known-insecure default now fails the process closed at startup; tests cover the raise and both accept paths.
- [x] **Contract is truth (P10):** enforces the "MUST override" the config docstring already stated.

## DCO sign-off

## Related issues

Refs #288 (API-04). The login/MFA rate-limiting item (API-03) from the same finding cluster is a separate follow-up PR.

### DIFF
--- a/api/app/config.py
+++ b/api/app/config.py
@@ -21,6 +21,11 @@ from pydantic_settings import BaseSettings, SettingsConfigDict
 
 LogLevel = Literal["debug", "info", "warning", "warn", "error", "critical"]
 
+# Published development default for the JWT signing secret. Refused at startup
+# outside dev mode (see ``assert_production_secrets``) so a deployment can never
+# silently ship with a public signing key.
+DEV_JWT_SECRET = "dev-jwt-secret-change-me"
+
 
 class Settings(BaseSettings):
     """Backend API configuration.
@@ -173,7 +178,7 @@ class Settings(BaseSettings):
 
     # ----- JWT (per ADR 0002 — backend owns auth) -----
     jwt_secret: str = Field(
-        default="dev-jwt-secret-change-me",
+        default=DEV_JWT_SECRET,
         description="Signing secret for JWT access and refresh tokens.",
     )
     jwt_access_token_ttl_seconds: int = Field(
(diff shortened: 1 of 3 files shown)


SaifAlYounan:
Superseded by #399 — this PR was auto-closed when I accidentally deleted my fork, and GitHub cannot
commit** (`84bbaf15a45c`); no content changed.

PR #293, opened 2026-07-07
Guard LLM provider base_url egress at build time (GW-04, Refs #288)

SaifAlYounan:
## What this PR does

Adds a build-time egress guard on LLM provider `base_url`s: `https` is always allowed, `http` only to a local host (Ollama/vLLM). A misconfigured provider now fails fast at startup instead of silently sending prompts in cleartext to a public host.

## Why

`Refs #288` — finding **GW-04 (Medium)**. `ProviderConfig.base_url` was validated only as `Field(min_length=1)`, and `build_adapter` handed it straight to the adapter with no scheme/host check. So the **crown-jewel prompt-egress path** could use a plaintext `http://` base_url or an internal/attacker-controlled address — while the **lower-trust tool path** is hardened (`build_tool_adapter` → `validate_base_url` → `validate_egress_target`: https-only, host allowlist, public-IP/DNS-rebind check). That asymmetry means a config mistake or tampering could silently exfiltrate every prompt.

## How

- New `gateway/app/providers/base_url_policy.py::validate_llm_base_url`, called at the top of `build_adapter` (main.py) before any adapter is constructed — mirroring how `build_tool_adapter` validates each tool provider at build time, so a bad `base_url` fails at startup.
- Policy is intentionally looser than the tool path because LLM providers differ: public providers (Anthropic/OpenAI/Azure/Vertex/Bedrock) **must** use `https`, but local providers legitimately speak plaintext `http` to a loopback address or a compose service name (`http://ollama:11434`, `http://vllm:8000/v1`). So: `https` any host; `http` only when the host is loopback / a private-or-link-local IP literal / `localhost` / a single-label service name. Everything else (plaintext to a public host, non-http(s) schemes, missing host) is refused.

## What this PR does *not* do

Doesn't add a per-provider host allowlist (the finding notes it as optional; the scheme/locality guard closes the cleartext-exfil path, and an allowlist can layer on later). Doesn't change the tool-egress path.

## Type of change

## Testing

<details>
<summary>Gates run locally (gateway/.venv)</summary>

- `pytest tests/test_llm_base_url_policy.py` → **17 passed** — accepts the shipped configs (anthropic/openai/vertex https, ollama/vllm/localhost/private-IP http); rejects http-to-public, http-to-public-IP, non-http schemes, and missing host; `build_adapter` raises on a plaintext-public `base_url` and still builds a normal local Ollama.
- `mypy app` (strict) → `Success: no issues found in 57 source files`.

</details>

- [x] **Egress (P1) / Fail restrictive (P4):** the sole audited egress boundary now refuses a cleartext-to-public or non-http(s) provider target and fails closed at startup; tests cover accept + reject paths.
- [x] **One governance path (P6):** applied at the same `build_adapter` build-time hook the tool path uses.

## DCO sign-off

## Related issues

Refs #288 (GW-04).

> `gateway/**` is a security-sensitive path — routes to `@legalquants/security` per `.github/CODEOWNERS`.

### DIFF
--- a/gateway/app/main.py
+++ b/gateway/app/main.py
@@ -74,6 +74,7 @@ from app.providers import (
     OpenAIAdapter,
     ProviderAdapter,
 )
+from app.providers.base_url_policy import validate_llm_base_url
 from app.providers.tool.base import ToolProviderAdapter
 from app.providers.tool.courtlistener import CourtListenerToolAdapter
 from app.providers.tool.echo import EchoToolAdapter
@@ -161,6 +162,11 @@ def build_adapter(provider: ProviderConfig) -> ProviderAdapter | None:
 
     if not provider.enabled:
         return None
+    # Egress guard (#288, GW-04): the prompt-carrying LLM path must not send
+    # cleartext to a public host or use a non-http(s) scheme. Validate before
+    # building any adapter so a bad base_url fails at startup — matching the
+    # build-time validation the tool path already does.
+    validate_llm_base_url(provider.base_url)
     if provider.type == "anthropic":
         return AnthropicAdapter.from_config(provider)
     if provider.type in ("openai", "openai_compatible"):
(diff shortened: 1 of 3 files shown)


SaifAlYounan:
Superseded by #400 — this PR was auto-closed when I accidentally deleted my fork, and GitHub cannot
commit** (`5acfd4dccc0a`); no content changed.

PR #317, opened 2026-07-24
fix(gateway): per-provider default_max_tokens for the Anthropic adapter

sergiomaldo:
## What this PR does

Adds a `default_max_tokens` field on the provider entry for the Anthropic adapter, used when an incoming OpenAI-format request omits `max_tokens`. It is read defensively from `ProviderConfig.model_extra` (same pattern as the existing `timeout_s`), falls back to the module constant `DEFAULT_MAX_TOKENS` (4096) when absent, and is clamped to `RequestValidationConfig.max_max_tokens` so a config typo can never inject a budget the gateway would reject on an explicit request. An explicit `request.max_tokens` always wins. Documented in `gateway.yaml.example`.

## Why

Anthropic Messages requires `max_tokens`, so the adapter injects a default when the caller omits it — previously hardwired to 4096 with no per-provider override, unlike `timeout_s` which is already operator-tunable. Frontends that never set `max_tokens` (Donna's api does not, and neither its UI nor its message schema exposes the field) silently truncate every reply at ~4K output tokens. Observed in production: a long template draft cut off mid-section, after which a review pass "correctly" flagged the never-generated sections as absent. Drafting workloads routinely need 8-16K output.

Found operating a production self-hosted legal deployment (a law firm running Donna v0.6.2 fully local, BYOK Anthropic through this gateway). This is finding 18 in that deployment's defect log.

## What this PR does *not* do

- Change the built-in 4096 fallback.
- Let the api layer pass `max_tokens` through, or surface `finish_reason=length` truncation as a visible UI warning — both worthwhile follow-ups.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds capability) — the fix is delivered as a small additive config field
- [ ] Breaking change (fix or feature that would change existing behavior)
- [ ] Documentation update
- [ ] Skill contribution (see attestation below)
- [ ] Refactor (no behavioral change)
- [ ] Test / CI / tooling

## Testing

- [x] Unit tests added/updated
- [ ] Integration tests added/updated (if applicable)
- [x] Manually tested against a real deployment (describe scenario below)
- [ ] Acceptance test plan updated (if changing skill behavior)

<details>
<summary>Manual testing notes (if applicable)</summary>

[...]

PR #318, opened 2026-07-24
fix(gateway): raise Anthropic default timeout to 300s; label client timeouts distinctly

### LINKED ISSUE #489: Proposal: sane defaults tuned by use case, not per-provider config — where does this break?
## Where we're leaning

**The Inference Gateway should ship sane defaults and let them be tuned by _use case_, not by per-provider operator config.** Tuning belongs at the call site — the part of the system that knows what a request is _for_ (a chat turn, a skill run, a long drafting job). The gateway's job is narrower: a sensible default when the caller says nothing, plus a hard ceiling nothing can exceed. Per-provider operator knobs, if they exist at all, are a narrow deployment escape hatch for troubleshooting — not the primary way to shape behaviour.

We're posting this as a position on purpose. We'd rather state where we're leaning and have you poke holes in it than ask an open "which way?" — especially if you've run a self-hosted, multi-provider LLM gateway in production.

## What that means concretely

Two open PRs prompted this, and under this position they land differently:

- **`max_tokens` is already tunable per request** — the gateway honours `request.max_tokens` and only falls back to a constant (4096) when it's absent. So "tune by use case" is mostly about making that fallback sane (likely per-model or per-tier, not one flat number) and having the use-case-aware callers — skills, playbooks, drafting — ask for what they need. On this view **#317's per-provider `default_max_tokens` knob is tuning on the wrong axis**: one provider serves every use case, so the budget should follow the task, not the provider.
- **Timeout has no per-request path today** — it's fixed per provider. So use-case timeout tuning would need new plumbing (per-request, or per-tier defaults). **#318's flat 60s→300s bump is a reasonable better-default, but it isn't use-case tuning** — it's the interim you'd keep until the real thing exists.
- **The cost guard stays:** the default is deliberately conservative (4096 exists to stop "accidentally enormous responses on requests that didn't specify a budget"), use cases opt _up_ explicitly, and a real gateway ceiling stops a runaway caller.

## Where this might break — challenge us

- **Do callers actually know their use case's budget?** Handing a skill author a `max_tokens` to guess is its own foot-gun. Is "tune at the call site" just relocating the config problem to a worse place?
- **Is "use case" even the right axis, or does the model dominate?** A model with an 8k output cap and one with 200k aren't interchangeable per task. If the real variable is the model, that looks more like config than use case.
- **Who stops a runaway if every caller sets its own budget?** Is a deployment-wide operator cap actually the load-bearing safety net — i.e. is the per-provider knob necessary after all, just for cost control rather than tuning?
- **Timeout: use case or provider latency?** A slow provider times out a fast use case; a fast provider serves a slow one fine. Maybe timeout is inherently provider-level and _shouldn't_ follow the use case.
- **Does a single "sane default" survive the model zoo?** If there's no one sane number and it has to be per-model or per-tier — is that just config under a different name?
- **Discoverability:** a self-hosted operator who wants "make all my drafts longer, everywhere" has nowhere to turn if tuning is only at the call site. Is that the operator knob's real justification?

## What happens next

This is a position put up for challenge, not a decision — but it is not open-ended. The path is deliberate:

1. **This issue gathers experience and objections** (please challenge the position above).
2. **It feeds a short ADR** that sets the policy once — the tuning axis (use case vs. per-provider vs. per-model/tier), what the gateway defaults, and whether an operator escape hatch exists at all.
3. **#317 and #318 are then resolved to match the ADR** — merged as-is, reworked to fit and merged, or closed with the reasoning recorded. Neither sits in limbo, and neither is closed for anything other than the policy this issue settles.

Both PRs stay open until the ADR lands, and neither is pre-judged here — both are appreciated contributions from @sergiomaldo that surfaced a real decision the project hadn't made.

Context for the code: the gateway is a first-party LQ.AI service (the security boundary; bring-your-own-keys; single-tenant per deployment); its config schema is `gateway/app/config.py` / `gateway.yaml.example`, documented in PRD §4.4. The per-provider config is our own, not inherited from the web frontend.

sergiomaldo:
## What this PR does

Two related fixes to the Anthropic adapter's client-timeout behavior:

1. Raises `DEFAULT_TIMEOUT_SECONDS` from 60.0 to 300.0. The per-provider `timeout_s` field still overrides for operators who want a tighter budget.
2. Client-side httpx timeouts (unary and streaming) now raise a `ProviderTimeoutError` subclass instead of the generic `ProviderNetworkError` (wire code and 503 mapping unchanged, mirroring the `ProviderModelNotFound` precedent), and the routing-log refusal-reason builder labels those rows `client_timeout:provider_unavailable` instead of `upstream_error:provider_unavailable`.

## Why

Frontier drafting responses of 4-16K output tokens routinely exceed 60s of generation time, so the old default aborted healthy long generations mid-flight. Worse, the abort was logged as `upstream_error:provider_unavailable` — indistinguishable from a genuine Anthropic outage. In production this looked exactly like a flaky provider: every large request (~12K tokens in / 4K out) died at 60-63s while small probes always passed; one successful run clocked 60,123 ms — a photo finish. On a typical single-BYOK self-host, the fallback chain has no other adapters, so each "blip" surfaced as a hard 503.

Found operating a production self-hosted legal deployment (a law firm running Donna v0.6.2 fully local, BYOK Anthropic through this gateway). Findings 15/15a in that deployment's defect log.

## What this PR does *not* do

- Add retry-with-backoff on the primary for retryable upstream statuses before walking fallbacks (finding 15's other half — a separate, larger change).
- Scale the timeout with `max_tokens`.
- Change the external error contract: wire code and HTTP 503 mapping are untouched; only the exception type and the routing-log label are more precise.

## Type of change

- [ ] New feature (non-breaking change that adds capability)
- [ ] Documentation update
- [ ] Skill contribution (see attestation below)
- [ ] Test / CI / tooling

## Testing

- [ ] Integration tests added/updated (if applicable)
- [x] Manually tested against a real deployment (describe scenario below)

<details>
<summary>Manual testing notes (if applicable)</summary>

Diagnosed on a live deployment by correlating routing-log `provider_unavailable` rows with request sizes: all failures at 60-63s on large requests, none on small ones; raising the provider `timeout_s` eliminated every occurrence. Unit tests pin the new 300s default, assert unary + streaming httpx timeouts raise `ProviderTimeoutError` with the client-timeout message, and assert the routing-log label is distinct from an upstream 5xx.

</details>

## Documentation

- [ ] Skill-authoring guide updated (if skill conventions change)

No listed doc surface covers adapter timeout defaults; `timeout_s` was already documented in `gateway.yaml.example`.

- [x] **Egress (P1):** n/a — no new outbound call; existing gateway adapter paths only.
- [x] **No raw payloads (P3):** the new routing-log label is a fixed string; no message content, args, or keys are logged.
- [x] **Closed set (P2):** n/a — no new capability.
- [x] **Fail restrictive (P4):** timeouts still fail the request with the same 503/wire code; the failure path (unary and streaming) is what the new tests exercise.
- [x] **Atomic audit (P5):** n/a — no state changes.
- [x] **One governance path (P6):** reuses the existing provider-error hierarchy and refusal-reason builder; `ProviderTimeoutError` subclasses the existing error exactly as `ProviderModelNotFound` does.
- [x] **Human gate (P7):** n/a — nothing destructive.
- [x] **Operator control (P8):** the per-provider `timeout_s` override is unchanged and still wins over the new default.
- [x] **User owns data (P9):** n/a — no matter context or residency impact.
- [x] **Contract is truth (P10):** wire codes and HTTP mappings unchanged, so no OpenAPI/schema update needed; the routing-log label change is documented here and in the commit message.

## DCO sign-off

- [x] All commits in this PR are signed off per the [DCO](../CONTRIBUTING.md#sign-off-developer-certificate-of-origin)

## Related issues

None filed upstream; findings 15/15a in the reporting deployment's defect log. Related: the `default_max_tokens` PR from the same log (finding 18) submitted separately.

## Reviewer notes

The 300s figure is a judgment call: large enough for 16K-token generations on current frontier models, small enough to still catch genuinely hung connections. If you would rather keep 60s and instead scale with `max_tokens`, happy to rework — the label fix (part 2) is valuable independently.

---

**Update 2026-09-13 (maintainer).** Scope widened per the review: `ProviderTimeoutError` is now raised by the OpenAI (incl. Azure) and Ollama adapters too, so the `client_timeout:` label is provider-independent. Adapter defaults are **600s** (Anthropic 300 → 600, OpenAI 60 → 600, Ollama 120 → 600): the 300s figure covered the 4–16K-token drafting turns measured here; #503's 370s document-production turn (@SaifAlYounan) needed more. The api's own gateway-client timeout is raised to 900s in #535 so it stays the loosest hop. Decision recorded in ADR 0027 (closes #489). Refs #503.

### DIFF
--- a/gateway/app/api/inference.py
+++ b/gateway/app/api/inference.py
@@ -88,6 +88,7 @@ from app.providers import (
     ProviderAuthError,
     ProviderHTTPError,
     ProviderNetworkError,
+    ProviderTimeoutError,
     ProviderUnsupportedError,
 )
 from app.router import (
@@ -270,9 +271,16 @@ def _failure_reason(error: ProviderAdapterError) -> str:
     plus the upstream HTTP status when known — the distinguishing signal
     an operator needs to tell a request rejection from a genuine outage
     when reading inference_routing_log rows.
+
+    A gateway-side timeout (:class:`ProviderTimeoutError`) is prefixed
+    ``client_timeout:`` rather than ``upstream_error:`` — we gave up
+    waiting, the provider may be healthy — so operators can tell a
+    too-tight ``timeout_s`` from a genuine provider outage.
     """
 
     code, _ = _classify_provider_error(error)
+    if isinstance(error, ProviderTimeoutError):
+        return f"client_timeout:{code}"
     if isinstance(error, ProviderHTTPError):
         return f"upstream_error:{code}:status={error.upstream_status}"
     return f"upstream_error:{code}"
--- a/gateway/app/providers/__init__.py
+++ b/gateway/app/providers/__init__.py
@@ -24,6 +24,7 @@ from app.providers.base import (
     ProviderHealth,
     ProviderHTTPError,
     ProviderNetworkError,
+    ProviderTimeoutError,
     ProviderUnsupportedError,
 )
 from app.providers.ollama import OllamaAdapter, ProviderModelNotFound
@@ -69,5 +70,6 @@ __all__ = [
     "ProviderHealth",
     "ProviderModelNotFound",
     "ProviderNetworkError",
+    "ProviderTimeoutError",
     "ProviderUnsupportedError",
 ]
--- a/gateway/app/providers/anthropic.py
+++ b/gateway/app/providers/anthropic.py
@@ -50,6 +50,7 @@ from app.providers.base import (
     ProviderHealth,
     ProviderHTTPError,
     ProviderNetworkError,
+    ProviderTimeoutError,
     ProviderUnsupportedError,
 )
 from app.providers.openai_schema import (
@@ -73,9 +74,15 @@ ANTHROPIC_API_VERSION = "2023-06-01"
 """Pinned Anthropic API version. Update deliberately; bump in lockstep
 with changes to the request/response translation below."""
 
-DEFAULT_TIMEOUT_SECONDS = 60.0
+DEFAULT_TIMEOUT_SECONDS = 300.0
 """Default per-request timeout. PRD §4.4 / gateway.yaml.example exposes
-``timeout_s`` on each provider; if absent we use this default."""
+``timeout_s`` on each provider; if absent we use this default.
+
+300s rather than the earlier 60s: frontier drafting responses of
+4-16K output tokens routinely exceed 60s of generation time, and a
+client-side timeout mid-generation surfaced as a provider outage even
+though Anthropic was healthy. Operators wanting a tighter budget set
+``timeout_s`` on the provider entry, which still overrides this."""
 
 DEFAULT_MAX_TOKENS = 4096
 """Anthropic Messages requires ``max_tokens``. When the OpenAI-format
@@ -313,6 +320,15 @@ class AnthropicAdapter(ProviderAdapter):
                 json=anthropic_body,
                 headers=self._auth_headers(),
             )
+        except httpx.TimeoutException as exc:
+            # Our own timeout elapsed — not an upstream failure. Raise the
+            # distinct subclass so the routing log can tell a too-tight
+            # timeout_s from an Anthropic outage.
+            raise ProviderTimeoutError(
+                f"timed out after {self._timeout:g}s waiting for Anthropic "
+                "(client-side timeout; raise timeout_s for long generations)",
+                details={"provider": self.name, "timeout_s": self._timeout},
+            ) from exc
         except httpx.HTTPError as exc:
             raise ProviderNetworkError(
                 f"failed to reach Anthropic: {type(exc).__name__}",
@@ -662,6 +678,14 @@ async def _anthropic_stream_iter(
                     continue
 
                 # ping / unknown -> ignore
+    except httpx.TimeoutException as exc:
+        # Our own timeout elapsed mid-stream — not an upstream failure.
+        # See the unary path for the rationale behind the distinct class.
+        raise ProviderTimeoutError(
+            "timed out waiting for Anthropic stream "
+            "(client-side timeout; raise timeout_s for long generations)",
+            details={"provider": provider_name},
+        ) from exc
     except httpx.HTTPError as exc:
         raise ProviderNetworkError(
             f"failed to stream from Anthropic: {type(exc).__name__}",
--- a/gateway/app/providers/anthropic.py
+++ b/gateway/app/providers/anthropic.py
@@ -74,15 +74,24 @@ ANTHROPIC_API_VERSION = "2023-06-01"
 """Pinned Anthropic API version. Update deliberately; bump in lockstep
 with changes to the request/response translation below."""
 
-DEFAULT_TIMEOUT_SECONDS = 300.0
+DEFAULT_TIMEOUT_SECONDS = 600.0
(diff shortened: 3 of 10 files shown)


houfu:
Hi thanks for filing this fix from observing your own runs. I wanted your elaboration on why you chose 300s instead of something like 100s or even 70s (based on your own observations). 

Some comments arising from automated review. Whether we take action on it depends on humans, but appreciate your comments if you have any:

* "Heads-up on end-to-end effect: the backend's own gateway client also defaults to 60s for unary calls (api/app/clients/gateway.py), so within the lq-ai stack this change moves the unary long-generation failure from a gateway 503 at ~60s to a backend 504 at ~60s — the labeling improvement still lands, but the 'long generations now succeed' claim only holds for streaming or direct-to-gateway callers (which may match your deployment). Any ratified default >60s needs a companion decision or a documented limitation on the backend hop."
* One consequence worth stating in the rationale: the router treats any ProviderNetworkError (including the new subclass) as fallback-eligible (gateway/app/router.py), so this change also moves 'time until fallback fires on a hung connection' from 60s to 300s. That may well be the right trade for long generations, but the docstring/PR rationale should acknowledge it so the number is a deliberate choice rather than a side effect. 
* "Nice mechanism — but as landed, only the Anthropic adapter raises ProviderTimeoutError; OpenAI (60s default, gateway/app/providers/openai.py, shared by Azure) and Ollama still surface client timeouts as upstream_error, so the routing-log label is provider-dependent for the same condition, and OpenAI keeps the exact 60s long-generation problem this PR fixes for Anthropic. Either extend the httpx.TimeoutException catch to the sibling adapters (they're structurally identical) or note the Anthropic-only scope in the ProviderTimeoutError docstring and let us file a follow-up for the rest."

houfu:
@sergiomaldo It seems further review also encountered this issue whereby other providers such as OpenAI and ollama are also 60s, when it should be 300s or similar. Are you willing to fix this? If you don't want to expand the scope of this PR, feel free to let me know.

houfu:
@sergiomaldo — I asked you two questions in July and August instead of telling you what we'd decided. Here's the decision, for #317, #318 and #535 together.

#489's position lands as ADR 0027: per-request `max_tokens` is the tuning axis; your per-provider knobs and the api timeout setting are the operator escape hatches it allows. All three PRs merge, in the order #318 → #504 → #535 → #317. Your `ProviderTimeoutError` / `client_timeout:` design is what the whole set generalises.

For this PR I'll push one signed follow-up commit onto your branch (it allows maintainer edits): the same `httpx.TimeoutException` catch in the OpenAI (incl. Azure) and Ollama adapters with tests, and adapter defaults at 600s rather than 300s — @SaifAlYounan measured a 370s document-production turn in #503, which 300s would still have cut off. Squash-merge keeps you as author.

#535: lands with the default raised to 900s and the connect leg capped at 10s. **One thing only you can do:** its commit has no `Signed-off-by:` — `git commit --amend -s` and force-push, and I'll rebase my follow-up on top.

#317: lands with the knob as filed; the clamp goes (it read a fresh `RequestValidationConfig()`, not the operator's value, and the ceiling is unenforced — now DE-392, credited to you); default becomes 16384.

Thanks for filing all three from a real deployment — they surfaced a decision the project hadn't made, and the ADR credits you for it.

PR #319, opened 2026-07-24
DE-358 item 5: encrypt chat pending-tool-call resume payload at rest (Fernet envelope)

SaifAlYounan:
## What / why

DE-358 item 5 (PRD §9): encryption-at-rest for the chat pending-tool-call resume payload. `chat_pending_tool_call.tool_call_args` / `.resume_state` hold the full conversation payload needed to resume the tool loop after a destructive-tool confirmation — the same sensitivity class as `messages.content`, but longer-lived than a request and sitting in a table whose whole purpose is replayable state. This PR envelope-encrypts both columns under `LQ_AI_MCP_MASTER_KEY`.

This implements the item-5 option surfaced for decision in the #275 thread (2026-07-07). That comment flagged the three costs; each is mitigated here:
- **schema/migration cost** → none: the Fernet token is stored *inside* the existing JSONB column as `{"_lq_ai_enc": 1, "token": "..."}`; no migration, and legacy plaintext rows pass through on read during the transition (pending rows carry a ~15-minute TTL, so they age out within minutes of deploy; fallback removable after one release).
- **key becomes a hard requirement** → only for the confirmation-gate persist path (and MCP OAuth, which already required it); the persist path fails closed with the module's operator-actionable missing-key error — never a plaintext fallback write.
- **decrypt failure mode** → fail-closed and explicit: resume decrypts after the committed atomic claim, so a tampered/wrong-key ciphertext denies the resume with a structured 409 and the row is consumed (never replayable). Error messages never carry ciphertext or payload contents.

Also threads `LQ_AI_MCP_MASTER_KEY` through `docker-compose.yml` / `docker-compose.release.yml` (api + arq-worker) and the env examples — it was not previously threaded anywhere despite the MCP OAuth store already needing it.

## Approach

- New `encrypt_payload_envelope` / `decrypt_payload_envelope` helpers in `api/app/security/encryption.py`, reusing `MCPTokenEncryptor` and its exception pair (no new key namespace, no new dependency).
- All three `ChatPendingToolCall` persist sites in `api/app/api/chats.py` (non-streaming gate, streaming gate, nested gate inside resume) write envelopes; the resume route decrypts both columns.
- Docs updated: PRD DE-358 item 5 status note; resume endpoint 409 description in `backend-openapi.yaml`.

Item 6 (api-side per-chat tier ceiling) stays tracking-only: its policy mapping (inference floors → egress ceiling, or an explicit `Project.max_egress_tier`) is the schema/PRD decision already routed to the committee in the #275 thread.

## Acceptance evidence

From `api/` (Python 3.12.13, throwaway pgvector container; dev stack untouched):
- `ruff format --check app tests` → 440 files already formatted
- `ruff check app tests` → All checks passed!
- `mypy app` → Success: no issues found in 190 source files
- `pytest -q` (full api suite) → **2635 passed, 1 skipped**
- New coverage: `api/tests/test_payload_envelope.py` (9 tests: round-trip, ciphertext shape/no plaintext in raw column, missing-key fail-closed both directions, legacy passthrough, wrong-key/tamper without leakage, malformed envelope); integration: gate persist asserts raw columns hold the envelope and no plaintext args/conversation text; `test_gate_persist_without_master_key_fails_closed`; `test_legacy_plaintext_row_still_resumes`; `test_tampered_envelope_denies_resume_with_structured_error` (409, row resolved, replay 409, no payload in body). Existing approve/deny/replay tests now exercise decrypt-and-resume by default.

## Operator notes

- **Worker rebuild:** api workers are touched (compose env). After pulling, rebuild/recreate `api` + `arq-worker` together so both pick up the env threading.
- Set `LQ_AI_MCP_MASTER_KEY` before using MCP OAuth or the chat destructive-tool confirmation gate; mint with `python -c 'from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())'`. Key rotation invalidates in-flight pending confirmations (15-minute TTL; users simply re-trigger).

Refs #275


SaifAlYounan:
Superseded by #403 — this PR was auto-closed when I accidentally deleted my fork, and GitHub cannot
commit** (`da3ddeac783e`); no content changed.

Issue #321, opened 2026-07-24
Coordinated roadmap sweep — claiming and building the unclaimed backlog (tracking)


SaifAlYounan:
## Purpose

As discussed across our weekly calls: the roadmap (~140 rows) has a large unclaimed backlog, and I'm going to work through it as a coordinated, sequential claim→build→PR sweep rather than let it sit. This issue is the tracking home for that sweep — one place to see what's being built, in what order, and to object or take an item for yourself.

## Claims discipline

The process is the one we already enforce: **every item gets its own claim issue** (`DE-XXX — <title>`) referencing this umbrella immediately before its build starts, then one PR per item. Live claim state is re-checked before each claim — anything claimed by someone else meanwhile is skipped.

**If you want any listed item, say so here (or on its claim issue) and I will skip it.** Nothing here reserves work ahead of its claim issue.

## Order of work

Stage 0 (done/in flight): PR #316 revision, DE-358 items (#284/#285/#319, refs #275), roadmap hygiene closing 14 verified-shipped rows (#320), this issue.

Then, in stage order (dependency-aware; each stage preceded by an OSS-survey research memo for its problem cluster):

1. **§1/§2 leftovers + tabular** — DE-370, DE-371, DE-374, DE-375, DE-376, 1.6 matter-intake UI, DE-322, DE-323, DE-324, DE-333, DE-288, DE-297→DE-298, DE-304, DE-308, DE-309/DE-310→DE-330, DE-331
2. **§3 mini-PRDs** — 3.1 (DE-086+235), 3.2 (DE-224), 3.4 (DE-222+223), 3.5 (DE-032+233), 3.6 (DE-225), 3.7 (DE-031+234), 3.8 (DE-263)
3. **§4 engineering discipline** — 4.1, 4.2, 4.3, DE-229, DE-230, DE-231→DE-239, DE-232, DE-240, DE-284, DE-291, DE-292 (+DE-318), DE-253, DE-252→DE-251, DE-250
4. **§5 observability** — DE-299, DE-300, DE-301, DE-302, DE-303, DE-314, DE-315, DE-317, DE-245, DE-246, DE-247
5. **§6 compliance & procurement** — 6.1, 6.2, 6.3 (DE-024), 6.4, 6.5, 6.6 (DE-113), 6.7 (DE-226), 6.10 (DE-115), 6.11 (DE-313)
[...]

PR #367, opened 2026-07-25
3.5: air-gap install verification CI (sealed-network Tier-1 smoke)

### LINKED ISSUE #366: 3.5 / DE-032+DE-233 — Air-gap install verification CI
I'd like to take this — building it now as part of the coordinated roadmap sweep (see umbrella issue #321). Build plan summary: a CI job that brings up the compose local profile with the Ollama model pre-pulled, seals the compose bridge with a deny-all egress rule, drives a login→chat→Tier-1 response smoke through the stack, and asserts zero non-RFC1918 packets (pcap artifact) plus routing-log evidence — with a negative test proving the seal trips on a cloud adapter. Research memo finding: no surveyed self-hosted project proves offline installability in CI; this is a differentiator. Operator runbook included.

SaifAlYounan:
## What / why

Roadmap 3.5 (DE-032 + DE-233): CI that *proves* the compose local profile installs and answers a Tier-1 chat with **zero external egress** — the air-gap OSS survey found no self-hosted project in the surveyed set (GitLab, Mattermost, Sentry, Ollama, k3s) proves offline installability in CI; they all do it by documentation. This makes the claim testable.

Part of the coordinated roadmap sweep (umbrella issue #321).

## Design

- **Seal:** iptables `LQ-AIRGAP` chain in `DOCKER-USER` on the compose bridge — RFC1918 + loopback RETURN, everything else (incl. 169.254/16) DROP. Chosen over `internal: true` because it certifies the *shipped* topology byte-identical (internal networks disable published ports, which the host-side driver needs); the internal-network variant is documented as the non-Linux local reproduction. DNS side channel documented honestly (resolution may succeed; connections die and are captured — CI is strictly more permissive than a real air gap, the safe direction).
- **Order matters:** images built + model pre-pulled, *then* seal, *then* the full stack boots **fresh under the seal** — so migrations, config seeding, and admin bootstrap are all proven offline, not just steady-state chat.
- **Journey:** first-run admin password from bootstrap logs → login → forced rotation → chat with `ollama-local/llama3.2:1b`, asserting content + `routed_provider=ollama-local` + tier 1, then routing-log assertions (≥1 tier-1 row, zero non-refused rows at any other tier). `gateway.yaml.example` ships unmodified; the smoke model is CPU-practical `llama3.2:1b` via raw passthrough (the pinned default is GPU-shaped) — the certified property is topology/routing, not model choice (documented).
- **Evidence:** tcpdump with a suspect-only filter → clean run uploads an **empty pcap**; the **negative control** (from inside the sealed gateway: direct-IP 1.1.1.1:443 connect + GET at the shipped Anthropic endpoint, both must fail, captured as a non-empty attempts pcap) closes the vacuous-pass hole. All evidence uploads as an artifact `if: always()`.
- **Triggers:** PRs touching compose/gateway-config/these scripts; weekly cron as the drift detector; manual dispatch. Deliberately not every PR (~30–45 min cold build; justified in-file).

## Runbook

`docs/security/air-gap-verification.md` (AI-drafted banner): what CI proves / does not prove, exact reproduction, per-component BOM, pre-fetch checklist (incl. the docling/EasyOCR ~700 MB first-ingestion download trap), and the honest **degradation table**: cloud tiers 2–5 (no silent cloud fallback by design), embedding alias → keyword-only retrieval unless repointed, citation judge repointing, CourtListener/GovInfo/EDGAR/EUR-Lex, Slack/Teams bridges, external MCP, OTel, budget email, ACME (→ internal CA, item 3.7), Word add-in CDN. **DE-386 filed** (PRD §9) for the k3s-style `docker save` release bundle follow-up.

## Honest verification split

Verified locally: shellcheck + bash -n on all scripts, workflow YAML parse, `docker compose --profile local config -q` renders, bridge-name derivation empirically checked on a scratch network, all wire shapes/log formats/column names confirmed against source. **CI-only (this job has NOT been run end-to-end):** the sealed boot itself and whether any service attempts egress at boot — if one does, the job fails, and that failure is the honest signal this exists to produce. First green run is the maintainer-visible acceptance.

## Notes for reviewers

`.github/workflows/**` + scripts = security-review routed. Runner assumption: GitHub-hosted ubuntu runners permit sudo iptables (standard); documented in the workflow.

Closes #366


SaifAlYounan:
**First live run analysis + fix pushed.** The run failed at `Assert zero egress` — and the failure log is exactly the signal this job exists to produce, plus one semantic bug of mine:

- **The seal held.** All 24 captured packets are outbound attempts that never connected: unanswered SYN retransmissions (to two Cloudflare IPs, a Google LB, and a DigitalOcean host — upstream phone-home/version-check behavior at boot) and FIN teardowns of connections opened during the pre-seal pull phase. Zero reply packets. The sealed boot, migrations, bootstrap, and the full Tier-1 chat smoke all **passed** before the assert.
- **My assert was too blunt:** it failed on *attempted* egress. In a true air gap those attempts simply time out — they're a property of upstream components, not a hole in the claim.

Fix (commit above): the capture now also records **breach** packets (non-private *source* = a reply got in = egress actually succeeded), and the assert hard-fails only on those. Blocked attempts are inventoried with per-container attribution into `sealed.attempts.txt` in the evidence artifact + the job summary, and the runbook documents the known sources — with the rule that any LQ.AI-authored component ever appearing in the inventory is a regression to treat as a bug. The certified claim is stated precisely: **zero successful egress, every attempt on the record.**

The push re-triggers the workflow (this PR touches its paths); the next run's summary will carry the attribution inventory.

SaifAlYounan:
Superseded by #427 — this PR was auto-closed when I accidentally deleted my fork, and GitHub cannot
commit** (`38977b7fde0e`); no content changed.

PR #291, opened 2026-07-07
Sanitize skill markdown before rendering (D-01 XSS, Refs #288)

SaifAlYounan:
## What this PR does

Sanitizes the skill markdown body in `SkillSourceView.svelte` with `DOMPurify` before it is injected via `{@html}`.

## Why

`Refs #288` — finding **D-01 (High, stored XSS)**. `SkillSourceView` computed `renderedMd = marked(contentMd ?? '', { breaks: true })` and injected it with `{@html}` and **no** sanitization — the one LQ-authored `{@html}` sink that skipped the DOMPurify convention the rest of the UI follows (`MessageBubble`, the inherited Open WebUI markdown pipeline). `marked@^9` removed its built-in sanitizer, so raw inline HTML passes straight through. Skill bodies are authored content — team-scope skills (a team admin's body renders for every member) and, for built-ins, the `skills/community` git submodule — so a body such as `` runs script in a viewer's authenticated session on opening the Source tab. Open WebUI stores the auth token in `localStorage`, so that is account/session takeover.

## How

`import DOMPurify from 'dompurify'` (already a dependency) and wrap the render: `DOMPurify.sanitize(marked(...) as string)` — identical to the pattern already used in `MessageBubble.svelte`.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)

## Testing

- [x] Manually reasoned against the sink; mirrors the audited-safe `MessageBubble` sanitization.

<details>
<summary>On automated tests</summary>

No vitest unit test was added: this repo's vitest runs in the **node** environment with no DOM (there is no `test`/jsdom config; tests exercise pure logic exported from `<script context="module">`, and DOM-rendering behavior is covered by **Cypress** per the house convention, e.g. `RefusalMessageBubble.test.ts`). `DOMPurify.sanitize` requires a DOM, so it cannot be exercised in that node harness. The change is a one-line reuse of the same DOMPurify call `MessageBubble` already relies on. `svelte-check` + prettier run in CI (`Web` job) — web deps were not installed in this sandbox. Happy to add a Cypress case if maintainers prefer.

</details>

## Documentation

- [x] n/a — no API/schema/behavior change beyond the sanitization.

## Transparency & governance invariants

- [x] **Fail restrictive (P4):** untrusted skill-body HTML is now stripped before rendering.

## DCO sign-off

- [x] All commits in this PR are signed off per the DCO.

## Related issues

Refs #288 (D-01).

[...]

PR #398, opened 2026-07-25
Sanitize skill markdown before rendering (D-01 XSS, Refs #288)

SaifAlYounan:
## What this PR does

Sanitizes the skill markdown body in `SkillSourceView.svelte` with `DOMPurify` before it is injected via `{@html}`.

## Why

`Refs #288` — finding **D-01 (High, stored XSS)**. `SkillSourceView` computed `renderedMd = marked(contentMd ?? '', { breaks: true })` and injected it with `{@html}` and **no** sanitization — the one LQ-authored `{@html}` sink that skipped the DOMPurify convention the rest of the UI follows (`MessageBubble`, the inherited Open WebUI markdown pipeline). `marked@^9` removed its built-in sanitizer, so raw inline HTML passes straight through. Skill bodies are authored content — team-scope skills (a team admin's body renders for every member) and, for built-ins, the `skills/community` git submodule — so a body such as `` runs script in a viewer's authenticated session on opening the Source tab. Open WebUI stores the auth token in `localStorage`, so that is account/session takeover.

## How

`import DOMPurify from 'dompurify'` (already a dependency) and wrap the render: `DOMPurify.sanitize(marked(...) as string)` — identical to the pattern already used in `MessageBubble.svelte`.

## Type of change

## Testing

- [x] Manually reasoned against the sink; mirrors the audited-safe `MessageBubble` sanitization.

<details>
<summary>On automated tests</summary>

No vitest unit test was added: this repo's vitest runs in the **node** environment with no DOM (there is no `test`/jsdom config; tests exercise pure logic exported from `<script context="module">`, and DOM-rendering behavior is covered by **Cypress** per the house convention, e.g. `RefusalMessageBubble.test.ts`). `DOMPurify.sanitize` requires a DOM, so it cannot be exercised in that node harness. The change is a one-line reuse of the same DOMPurify call `MessageBubble` already relies on. `svelte-check` + prettier run in CI (`Web` job) — web deps were not installed in this sandbox. Happy to add a Cypress case if maintainers prefer.

</details>

## Documentation

- [x] n/a — no API/schema/behavior change beyond the sanitization.

- [x] **Fail restrictive (P4):** untrusted skill-body HTML is now stripped before rendering.

## DCO sign-off

## Related issues

Refs #288 (D-01).

---

_Re-raised after I accidentally deleted my fork, which auto-closed the original PR. **Supersedes #291** — same head commit (`75aacc1e802d`), no content change. GitHub cannot reopen a cross-repo PR once its head repository is gone, hence the new number._

### DIFF
--- a/web/cypress/e2e/wave-b-surfaces.cy.ts
+++ b/web/cypress/e2e/wave-b-surfaces.cy.ts
@@ -10,6 +10,7 @@
  *   5. /lq-ai/admin/developer renders the four developer-support cards
  *   6. ✨ Enhance Prompt button opens the expansion panel (or error state)
  *   7. /lq-ai/skills/[id] detail page renders SkillDetailTabs; tab switching works
+ *   8. Source tab does not execute a hostile skill body (D-01 XSS regression)
  *
  * Run requires a live stack:
  *   docker compose up -d
@@ -17,6 +18,12 @@
  *   (note the printed password; export LQAI_ADMIN_PASSWORD or update env)
  *   cd web && npx cypress run --spec 'cypress/e2e/wave-b-surfaces.cy.ts'
  */
+
+import { getBearerToken } from '../support/lq-ai-helpers';
+
+/** Direct API base — the SvelteKit web container has no POST proxy for user-skills routes. */
+const API_BASE = () => Cypress.env('LQAI_API_BASE') ?? 'http://localhost:8000';
+
 describe('Wave B v2 — new surfaces', () => {
   beforeEach(() => {
     cy.visit('/lq-ai/login');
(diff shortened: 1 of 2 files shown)


houfu (changes requested):
Could you add a Cypress case that opens the Source tab on a skill whose body carries a hostile payload and asserts nothing executes? Our contributing guide asks for a regression test on every bug fix, and `web/cypress/e2e/wave-b-surfaces.cy.ts` already has a test that opens that tab — extending it is the smallest path. Cypress is already set up in the repo, so this needs no new tooling.

SaifAlYounan:
Thanks @houfu — regression test added as test 8 in `web/cypress/e2e/wave-b-surfaces.cy.ts`,
pushed as `4f7d2c10`. Extended the existing spec as you suggested; no new tooling, and it stays
inside the deterministic track (no LLM round-trip, no provider key).

It seeds a real user-skill through the API rather than stubbing the response, because the
server stores the body verbatim (`api/app/api/skills.py` returns `row.body` unchanged). That
way the test covers the whole path and would still catch the bug if the storage layer changed
underneath the component. Payloads are an eager-loading tag with an event handler, a raw
`<script>`, and a `javascript:` URL. It asserts the body actually rendered *before* asserting
on absences — otherwise a page that failed to load would satisfy every "should not exist" and
pass while proving nothing — then that none of the three survived and that nothing executed.

**One honest caveat: I have not been able to execute this spec.** Cypress 13.17 will not start
on my machine — it crashes in its config loader with an empty error object, identically on node
20 and node 22, with both a TS and a plain-JS config, and after a forced binary reinstall. So
the spec is hand-reviewed and eslint-clean, but unrun. Its first real execution will be the
nightly deterministic track. If that shows a selector problem, it's mine and I'll fix it
promptly.

What I *could* verify, and did, is the security claim itself — running the exact shipped
`marked` and `dompurify` versions from `web/node_modules` over the same payload:

```
BEFORE:  <img src=x onerror="...">  <script>...</script>  <a href="javascript:...">
AFTER:   <img src="x">              (removed)            <a>click me</a>
```

All three neutralised, benign content preserved. I also ran the test's own assertions against
the *pre*-sanitization output and confirmed all three fail there — so it is a real regression
test, not one that would pass with or without the fix.

houfu (approved):
LGTM

PR #400, opened 2026-07-25
Guard LLM provider base_url egress at build time (GW-04, Refs #288)

SaifAlYounan:
## What this PR does

Adds a build-time egress guard on LLM provider `base_url`s: `https` is always allowed, `http` only to a local host (Ollama/vLLM). A misconfigured provider now fails fast at startup instead of silently sending prompts in cleartext to a public host.

## Why

`Refs #288` — finding **GW-04 (Medium)**. `ProviderConfig.base_url` was validated only as `Field(min_length=1)`, and `build_adapter` handed it straight to the adapter with no scheme/host check. So the **crown-jewel prompt-egress path** could use a plaintext `http://` base_url or an internal/attacker-controlled address — while the **lower-trust tool path** is hardened (`build_tool_adapter` → `validate_base_url` → `validate_egress_target`: https-only, host allowlist, public-IP/DNS-rebind check). That asymmetry means a config mistake or tampering could silently exfiltrate every prompt.

## How

- New `gateway/app/providers/base_url_policy.py::validate_llm_base_url`, called at the top of `build_adapter` (main.py) before any adapter is constructed — mirroring how `build_tool_adapter` validates each tool provider at build time, so a bad `base_url` fails at startup.
- Policy is intentionally looser than the tool path because LLM providers differ: public providers (Anthropic/OpenAI/Azure/Vertex/Bedrock) **must** use `https`, but local providers legitimately speak plaintext `http` to a loopback address or a compose service name (`http://ollama:11434`, `http://vllm:8000/v1`). So: `https` any host; `http` only when the host is loopback / a private-or-link-local IP literal / `localhost` / a single-label service name. Everything else (plaintext to a public host, non-http(s) schemes, missing host) is refused.

## What this PR does *not* do

Doesn't add a per-provider host allowlist (the finding notes it as optional; the scheme/locality guard closes the cleartext-exfil path, and an allowlist can layer on later). Doesn't change the tool-egress path.

## Type of change

- [x] Bug fix (fixes an issue)
- [x] Breaking change (a deployment whose provider `base_url` violates the egress policy — e.g. a mistyped `OLLAMA_BASE_URL` pointing at a public host over plaintext `http` — no longer starts with that provider skipped; the gateway refuses to boot, deliberately, naming the provider. The tradeoffs table in the description is the upgrade note.)

## Testing

<details>
<summary>Gates run locally (gateway/.venv)</summary>

- `pytest tests/test_llm_base_url_policy.py` → **17 passed** — accepts the shipped configs (anthropic/openai/vertex https, ollama/vllm/localhost/private-IP http); rejects http-to-public, http-to-public-IP, non-http schemes, and missing host; `build_adapter` raises on a plaintext-public `base_url` and still builds a normal local Ollama.
- `mypy app` (strict) → `Success: no issues found in 57 source files`.

</details>

- [x] **Egress (P1) / Fail restrictive (P4):** the sole audited egress boundary now refuses a cleartext-to-public or non-http(s) provider target and fails closed at startup; tests cover accept + reject paths.
- [x] **One governance path (P6):** applied at the same `build_adapter` build-time hook the tool path uses.

## DCO sign-off

## Related issues

Refs #288 (GW-04).

> `gateway/**` is a security-sensitive path — routes to `@legalquants/security` per `.github/CODEOWNERS`.

---

_Re-raised after I accidentally deleted my fork, which auto-closed the original PR. **Supersedes #293** — same head commit (`5acfd4dccc0a`), no content change. GitHub cannot reopen a cross-repo PR once its head repository is gone, hence the new number._

---

## Update — explicit allowlist (review feedback, 2026-07-28)

Per @houfu's review, the plaintext-`http` decision is now driven by two named constants
an operator can read in the file, instead of Python's `ip.is_global`:

```python
_LOCAL_HOSTS    = {"localhost", "host.docker.internal", "ollama", "vllm"}
_LOCAL_NETWORKS = 127.0.0.0/8, ::1/128, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, fc00::/7
```

**This also fixes a startup regression the first version introduced.** The single-label
heuristic treated any dotted name as a public FQDN, so it refused
`http://host.docker.internal` — the value `.env.example` ships as the default
`OLLAMA_BASE_URL`, and the only route to Ollama in `docker-compose.release.yml`, which
has no `ollama` service. `build_adapter` raises inside `lifespan`, which catches only
`ValueError`, so the gateway failed to start on the most common documented configuration.
Reproduced and confirmed fixed.

**Behaviour change worth reviewing deliberately.** The explicit network list is narrower
than `not is_global`. Plaintext `http` to these is now **refused** where it was previously
allowed:

| Range | Note |
|---|---|
| `169.254.0.0/16` | link-local — carries the cloud metadata endpoint (`169.254.169.254`) |
| `100.64.0.0/10` | CGNAT — where Tailscale addresses live |
| `0.0.0.0`, `fe80::/10` | not local inference targets |

An unlisted single-label service name (e.g. `http://nginx:8080`) is also refused now. The
intended remedy is to add the name to `_LOCAL_HOSTS` rather than loosen the rule. I took
the list exactly as proposed rather than widening it; if any operator recipe needs the
CGNAT range for a tailnet-hosted Ollama, that is a one-line addition and worth a follow-up.

Tests go 17 → 20: accept `host.docker.internal`, reject the IMDS address and an unlisted
single-label name so a later edit cannot silently restore the heuristic. Full gateway
suite green (790 passed, 3 skipped), `ruff`, `ruff format --check` and `mypy --strict` clean.

### DIFF
--- a/gateway/app/providers/base_url_policy.py
+++ b/gateway/app/providers/base_url_policy.py
@@ -1,0 +1,70 @@
+"""Egress policy for LLM provider ``base_url``s (#288, GW-04).
+
+The tool path (:mod:`app.providers.tool.egress`) enforces a strict
+https-only, host-allowlisted, public-IP egress guard. The LLM provider path
+needs a slightly looser but still-safe rule: public providers (Anthropic,
+OpenAI, Azure, Vertex, Bedrock) MUST use ``https``, but local providers
+(Ollama, vLLM) legitimately speak plaintext ``http`` to a loopback address
+or a compose service name (``http://ollama:11434``, ``http://vllm:8000/v1``).
+
+Before this guard ``ProviderConfig.base_url`` was validated only as
+``min_length=1``, so a misconfiguration or config tampering could point the
+crown-jewel prompt-egress path at ``http://attacker.example`` or an internal
+address in cleartext — the asymmetry the audit flagged versus the hardened
+tool path. This guard is applied at adapter build time so a bad ``base_url``
+fails fast at startup.
+"""
+
+from __future__ import annotations
+
+import ipaddress
+from urllib.parse import urlparse
+
+
+class ProviderEgressRefused(Exception):
+    """Raised when a provider ``base_url`` violates LLM egress policy."""
+
+    def __init__(self, reason: str) -> None:
+        super().__init__(reason)
+        self.reason = reason
+
+
+def _is_local_host(host: str) -> bool:
+    """True when plaintext ``http`` egress to ``host`` is acceptable.
+
+    Acceptable local targets: a loopback/private/link-local/CGNAT IP literal,
+    ``localhost``, or a single-label hostname (no dot) — i.e. a container /
+    compose service name such as ``ollama`` or ``vllm`` that is not a routable
+    public host. A dotted FQDN (``api.openai.com``) is treated as public.
+    """
+    if host.lower() == "localhost":
+        return True
+    try:
+        ip = ipaddress.ip_address(host)
+    except ValueError:
+        # Not an IP literal: a single-label name is a private service name;
+        # a dotted (or IPv6-bracketed) name is a public FQDN.
+        return "." not in host and ":" not in host
+    return not ip.is_global
+
+
+def validate_llm_base_url(base_url: str) -> None:
+    """Validate an LLM provider ``base_url`` against egress policy.
+
+    ``https`` is always allowed; ``http`` is allowed only to a local host
+    (Ollama/vLLM). Any other scheme, a missing host, or plaintext ``http`` to
+    a public host raises :class:`ProviderEgressRefused`.
+    """
+    parsed = urlparse(base_url)
+    if parsed.scheme not in ("http", "https"):
+        raise ProviderEgressRefused(
+            f"provider base_url must use http or https, got scheme {parsed.scheme!r}"
+        )
+    host = parsed.hostname
+    if not host:
+        raise ProviderEgressRefused("provider base_url has no host")
+    if parsed.scheme == "http" and not _is_local_host(host):
+        raise ProviderEgressRefused(
+            "plaintext http base_url is only permitted for local providers "
+            f"(loopback/private/service name); public host {host!r} must use https"
+        )
--- a/gateway/app/providers/base_url_policy.py
+++ b/gateway/app/providers/base_url_policy.py
@@ -4,8 +4,14 @@ The tool path (:mod:`app.providers.tool.egress`) enforces a strict
 https-only, host-allowlisted, public-IP egress guard. The LLM provider path
 needs a slightly looser but still-safe rule: public providers (Anthropic,
 OpenAI, Azure, Vertex, Bedrock) MUST use ``https``, but local providers
-(Ollama, vLLM) legitimately speak plaintext ``http`` to a loopback address
-or a compose service name (``http://ollama:11434``, ``http://vllm:8000/v1``).
+(Ollama, vLLM) legitimately speak plaintext ``http`` to a loopback address,
+a private network, or one of a few named local targets
+(``http://ollama:11434``, ``http://host.docker.internal:11434``).
+
+The named hosts and networks are spelled out as explicit module constants
+below rather than derived from :attr:`ipaddress.ip_address.is_global`, so an
+operator can see exactly what plaintext egress is permitted without leaving
+this file.
 
 Before this guard ``ProviderConfig.base_url`` was validated only as
 ``min_length=1``, so a misconfiguration or config tampering could point the
@@ -29,31 +35,60 @@ class ProviderEgressRefused(Exception):
         self.reason = reason
 
 
+# Hostnames permitted as plaintext ``http`` targets. Every local-inference
+# host the project documents is here: ``ollama`` is the Compose service
+# (docker-compose.yml), ``vllm`` is the disabled-by-default service in
+# gateway.yaml.example, and ``host.docker.internal`` is the host bridge that
+# .env.example ships as the default OLLAMA_BASE_URL (the release Compose file
+# has no ollama service, so that is the release path). Add a name here rather
+# than loosening the rule.
+_LOCAL_HOSTS = frozenset({"localhost", "host.docker.internal", "ollama", "vllm"})
+
+# Networks permitted as plaintext ``http`` targets: loopback, RFC1918 private
+# space, and IPv6 unique-local. Deliberately narrower than "not is_global" —
+# link-local (169.254.0.0/16, which carries the cloud metadata endpoint),
+# CGNAT (100.64.0.0/10) and 0.0.0.0 are NOT local inference targets and must
+# not receive prompts in cleartext.
+_LOCAL_NETWORKS = tuple(
+    ipaddress.ip_network(n)
+    for n in (
+        "127.0.0.0/8",
+        "::1/128",
+        "10.0.0.0/8",
+        "172.16.0.0/12",
+        "192.168.0.0/16",
+        "fc00::/7",
+    )
+)
+
+
 def _is_local_host(host: str) -> bool:
     """True when plaintext ``http`` egress to ``host`` is acceptable.
 
-    Acceptable local targets: a loopback/private/link-local/CGNAT IP literal,
-    ``localhost``, or a single-label hostname (no dot) — i.e. a container /
-    compose service name such as ``ollama`` or ``vllm`` that is not a routable
-    public host. A dotted FQDN (``api.openai.com``) is treated as public.
+    Acceptable local targets: a name in :data:`_LOCAL_HOSTS`, or an IP literal
+    inside one of :data:`_LOCAL_NETWORKS`. Anything else — a public FQDN such
+    as ``api.openai.com``, an unlisted service name, or a routable IP — is
+    treated as public and must use ``https``.
     """
-    if host.lower() == "localhost":
+    if host.lower() in _LOCAL_HOSTS:
         return True
     try:
         ip = ipaddress.ip_address(host)
     except ValueError:
-        # Not an IP literal: a single-label name is a private service name;
-        # a dotted (or IPv6-bracketed) name is a public FQDN.
-        return "." not in host and ":" not in host
-    return not ip.is_global
+        # Not an IP literal and not an allowlisted name.
+        return False
+    # A v4 address is never "in" a v6 network (and vice versa); CPython returns
+    # False on a version mismatch rather than raising, so no version guard.
+    return any(ip in network for network in _LOCAL_NETWORKS)
 
 
 def validate_llm_base_url(base_url: str) -> None:
     """Validate an LLM provider ``base_url`` against egress policy.
 
-    ``https`` is always allowed; ``http`` is allowed only to a local host
-    (Ollama/vLLM). Any other scheme, a missing host, or plaintext ``http`` to
-    a public host raises :class:`ProviderEgressRefused`.
+    ``https`` is always allowed; ``http`` is allowed only to a host in
+    :data:`_LOCAL_HOSTS` or an IP inside :data:`_LOCAL_NETWORKS`. Any other
+    scheme, a missing host, or plaintext ``http`` to a public host raises
+    :class:`ProviderEgressRefused`.
     """
     parsed = urlparse(base_url)
     if parsed.scheme not in ("http", "https"):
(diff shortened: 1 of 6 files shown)


gateway/app/providers/base_url_policy.py
```diff
@@ -0,0 +1,70 @@
...
+
+def _is_local_host(host: str) -> bool:
+    """True when plaintext ``http`` egress to ``host`` is acceptable.
+
+    Acceptable local targets: a loopback/private/link-local/CGNAT IP literal,
+    ``localhost``, or a single-label hostname (no dot) — i.e. a container /
+    compose service name such as ``ollama`` or ``vllm`` that is not a routable
+    public host. A dotted FQDN (``api.openai.com``) is treated as public.
+    """
+    if host.lower() == "localhost":
```
houfu: We can specify a list of text names instead of limiting to only localhost. 

Examples like: 
* host.docker.internal
* ollama
* vllm

Using a list of names will also make it more transparent what is in the allowlist.

gateway/app/providers/base_url_policy.py
```diff
@@ -0,0 +1,70 @@
...
+        return True
+    try:
+        ip = ipaddress.ip_address(host)
+    except ValueError:
+        # Not an IP literal: a single-label name is a private service name;
+        # a dotted (or IPv6-bracketed) name is a public FQDN.
+        return "." not in host and ":" not in host
+    return not ip.is_global
```
houfu: Instead of relying on python's implementation of `ip.is_global`, let's use a list so that administrators can see what is allowed without leaving the file.

```python
_LOCAL_HOSTS = frozenset({"localhost", "host.docker.internal", "ollama", "vllm"})
_LOCAL_NETWORKS = tuple(
    ipaddress.ip_network(n)
    for n in ("127.0.0.0/8", "::1/128", "10.0.0.0/8", "172.16.0.0/12", "192.168.0.0/16", "fc00::/7")
)
```

gateway/tests/test_llm_base_url_policy.py
```diff
@@ -0,0 +1,81 @@
...
+    "url",
+    [
+        "http://api.openai.com/v1",  # plaintext to a public host
+        "http://attacker.example/v1",
+        "http://8.8.8.8/v1",  # plaintext to a public IP
+        "ftp://api.openai.com",  # unsupported scheme
+        "file:///etc/passwd",
+        "https://",  # no host
+    ],
+)
```
houfu, line 56: Test should include http://host.docker.internal

houfu (changes requested):
The code works but my recommendation is to be explicit within the file about which hosts are allowed or disallowed. 

http://host.docker.internal should be included as the name is specified in docs and examples

SaifAlYounan:
Thanks @houfu — all three points are in, pushed as `f554faed`.

`_LOCAL_HOSTS` and `_LOCAL_NETWORKS` are now module constants shaped as you proposed, and
`_is_local_host` reads from them instead of `ip.is_global`. `host.docker.internal` is in the
list and covered by a test.

**Your `host.docker.internal` point was more serious than a missing test case, so flagging it
explicitly.** The single-label heuristic treated any dotted name as a public FQDN, so it
refused `http://host.docker.internal` — which is the value `.env.example` ships as the default
`OLLAMA_BASE_URL`, and the only route to Ollama in `docker-compose.release.yml`, which has no
`ollama` service. `build_adapter` raises inside `lifespan`, and `lifespan` catches only
`ValueError`, so **the gateway failed to start on the most common documented configuration.**
I reproduced it and confirmed the fix clears it. Good catch.

**One thing to review deliberately: the explicit network list is narrower than `not is_global`.**
Plaintext `http` to these was allowed before and is refused now:

| Range | Note |
|---|---|
| `169.254.0.0/16` | link-local — carries the cloud metadata endpoint `169.254.169.254` |
| `100.64.0.0/10` | CGNAT — where Tailscale addresses live |
| `0.0.0.0`, `fe80::/10` | not local inference targets |

An unlisted single-label name (`http://nginx:8080`) is refused now too. I took your list
exactly rather than widening it, on the view that an operator should add a name to
`_LOCAL_HOSTS` rather than have the rule guess. The one I'd flag is CGNAT: there is a
`deploy/caddy-tailscale/` recipe in-tree, so if anyone reaches a tailnet-hosted Ollama that
way it would now be refused. No `100.64.*` base_url exists anywhere in the repo, so I left it
out — happy to add it if you'd rather.

Tests 17 → 20: accept `host.docker.internal`, reject the IMDS address and an unlisted
single-label name so a later edit can't quietly restore the heuristic. Full gateway suite
green (790 passed, 3 skipped); `ruff`, `ruff format --check`, `mypy --strict` clean.

Two follow-ups came out of this, both split out rather than folded in here:

- **#443** — `OLLAMA_BASE_URL` isn't pinned in `example_env`, so the gateway suite inherits the
  developer's shell. A developer who followed `.env.example` gets a failure that reproduces on
  their machine and not the next one. One line, independent of this PR.
- **`ProviderEgressRefused` is a bare `Exception`** and both `build_adapter` call sites catch
  only `ValueError`, so a BYOK provider-key write against a refused `base_url` returns an
  unhandled 500. Fixed and tested on `SaifAlYounan:fix/provider-egress-refused-typed-error`,
  but held rather than opened as a PR: the class doesn't exist on `main` yet, so it can only be
  stacked on this branch, and GitHub won't take a fork branch as a base. I'll open it against
  `main` the moment this merges. Startup behaviour stays fail-closed there — I deliberately did
  not widen the `lifespan` catch.

houfu (changes requested):
Thanks for turning the review round so thoroughly — the named constants read much better than `ip.is_global`, and catching the `host.docker.internal` regression was a genuinely important find. That one would have broken the most common documented setup, and you spotted it, reproduced it, and covered it with a test.

You also flagged the exception-type problem yourself and held a fix for it. We'd like to take you up on that, with one change of direction — and the reason is a design call we've now made, not anything wrong with what you wrote.

**The decision: a refused `base_url` should stop the gateway starting, deliberately.**

We looked at what happens if a refused provider is merely skipped. The router walks the
primary→fallback chain and skips any candidate with no adapter (`gateway/app/router.py:664-672`), so an operator running a local Ollama primary with a cloud fallback who mistypes `OLLAMA_BASE_URL` would get a gateway that starts cleanly and quietly sends those prompts to the cloud provider instead. For a self-hosted deployment where the operator chooses where inference happens, that's the outcome we most want to avoid — so the loud failure is the behaviour we want to keep.

That is what the PR does today, but it happens because `ProviderEgressRefused` *misses* the `except ValueError` rather than because anything says so. Two changes would make it deliberate:

**1. Catch it explicitly in `lifespan`**, so a later refactor can't quietly turn it back into a
skip, and so the operator gets a message naming the provider rather than a bare traceback. In `gateway/app/main.py`, an arm ahead of the existing `except ValueError`:

```python
        try:
            adapter = build_adapter(provider)
        except ProviderEgressRefused as exc:
            # Deliberate: an egress-policy violation is fatal at startup.
            # Skipping the provider would let the router fall through to the
            # next candidate in the chain, silently sending prompts somewhere
            # the operator did not choose.
            logger.error(
                "refusing to start: provider %r (type=%s) has a base_url that "
                "violates LLM egress policy: %s",
                provider.name,
                provider.type,
                exc.reason,
            )
            raise
        except ValueError as exc:
```

with the class added to the import this PR already introduces:

```python
from app.providers.base_url_policy import ProviderEgressRefused, validate_llm_base_url
```

Putting the new arm first keeps it correct even if the class's base ever changes.

**2. Handle it on the BYOK path**, so the key-write returns a structured error instead of an unhandled 500 — the half you'd already identified. `apply_provider_key` (`gateway/app/provider_keys.py:325`) catches only `ValueError`, and by the time `build_adapter` runs the encrypted key has already been written and the config reloaded, so the request currently 500s *after* a successful write. An arm before the existing one:

```python
        try:
            new_adapter = build_adapter(provider)
        except ProviderEgressRefused as exc:
            # The key was written, but this provider's base_url is refused by
            # egress policy; report it rather than 500 after a successful write.
            raise ProviderKeyMutationError(
                f"provider {provider_name!r} base_url violates egress policy: {exc.reason}",
                http_status=409,
            ) from exc
        except ValueError:
```

`ProviderKeyMutationError` comes from `app.config_writer` and already carries the status through to the error envelope. It subclasses `ValueError`, so the new arm does need to sit first. 409 seemed the closest fit, since the stored config conflicts with policy — if you'd rather it were 400, that's fine by us.

Two smaller things would help it land, since this makes the narrowing an operator-visible upgrade behaviour: ticking **Breaking change** alongside Bug fix (the table you already wrote is exactly the upgrade note that belongs with it), and a line in `gateway.yaml.example` beside the local-provider block naming the rule, so an operator who meets the startup failure has somewhere to read why.

One thing you can take off your list: we looked at `deploy/caddy-tailscale/` and it's an
inbound-only recipe — its `.env.example` ships `host.docker.internal`, and there's no `100.64.*` `base_url` anywhere in it. So leaving CGNAT out breaks nothing shipped today, and taking the list as proposed rather than widening it was the right call.

**So the one next step: add those two `except ProviderEgressRefused` arms and push.** Everything else above is optional polish.

SaifAlYounan:
Both arms are in — pushed as `e8ee9ecc`.

**Lifespan** (`gateway/app/main.py`): the `except ProviderEgressRefused` arm sits ahead of the `ValueError` skip, logs the offending provider by name, and re-raises — your comment on why the loud failure is the behaviour we want is in the code verbatim, so the next refactor reads the reasoning where it matters.

**BYOK** (`gateway/app/provider_keys.py`): `ProviderEgressRefused` now maps to `ProviderKeyMutationError` with `http_status=409` before the `ValueError` arm (which it must, since `ProviderKeyMutationError` subclasses `ValueError` — noted in a comment there too). 409 kept as you proposed.

Also in the push:

- **Breaking change** is now ticked in the description, with the upgrade note pointing at the tradeoffs table.
- One doc block in `gateway.yaml.example` beside the local-provider entry names the fatal-at-startup rule and points a stranded operator at the usual culprit (a mistyped `OLLAMA_BASE_URL`).
- Two regression tests: the lifespan one reproduces exactly that operator story (example config + `OLLAMA_BASE_URL=http://attacker.example/v1` → startup raises), and the BYOK one pins the 409 mapping and asserts no adapter is swapped in on refusal. Gateway suite 792 passed, `ruff` + `mypy --strict` clean.

And thanks for checking `deploy/caddy-tailscale/` — taking CGNAT off the list was the call I was least sure of, so the confirmation is genuinely useful.

*Posted by AlexiOS, Alexios's personal AI agent, on his instruction.*

houfu (approved):
**This is complete and I'm merging it.** Three notes so nothing lands on you afterwards:

**1. Don't open `fix/provider-egress-refused-typed-error`.** You held it to raise against `main` once this merged — it's superseded. Re-basing `ProviderEgressRefused` onto `ValueError` and catching it explicitly are alternative fixes for the same defect, and we took the explicit-arm route deliberately: it survives someone changing the exception's base later, which the typed version doesn't. You can drop that branch.

**2. The DCO question is resolved, and it resolved in your favour.** `f554faed` is authored as `SaifAlYounan` but signed off as `Alexios Kirillov`, which I'd flagged to check. GitHub resolves `5acfd4dc` — authored from `alexiosvdskirillov@gmail.com` — to your account, so both identities are verified on the same GitHub user and the sign-off certifies for the same person. Nothing to fix and no rewrite needed. If it's easy on your end, pointing `git config user.email` at one of the two would save the next reviewer the same lookup, but that's a convenience, not a request.

**3. The remaining items from review are mine, not yours.** For the record so they're not orphaned:

- **The ADR.** This PR is the first recorded egress policy for the inference path — ADR 0014 D2 is scoped to *tool* egress on its own wording, so the divergence needed a decision record. I'm writing it as ADR 0026, crediting the policy and the allowlist to you. It also records the two calls made in this thread (enumeration over `is_global`; fatal-at-startup) which currently live only in a code comment.
- **The DNS question you effectively raised** — the allowlist trusts its names without resolving them, unlike the tool path — goes into that ADR as an explicitly open question for `@legalquants/security`, with both positions written up and neither ranked. If it lands on "resolve", that's a follow-up against `main`, and merging this is what unblocks it.
- **`model_discovery.py`** reaches provider `base_url`s from the live config snapshot outside this guard, sending the API key on the Anthropic and OpenAI paths. That's out of scope here — this PR does what GW-04 asked for — but it means the "sole audited egress boundary" framing isn't true yet, so I'm filing it separately rather than letting it ride on your PR's description.
- **#443** stands on its own and I'll take it next.

On the substance: catching the `host.docker.internal` regression yourself was the most valuable thing in this whole review round. It would have broken the most common documented setup, it was invisible to CI, and you found it, reproduced it, and pinned it with a test — after the first revision had already been reviewed. That's the part that's hard to teach.

Merging.

PR #295, opened 2026-07-07
Scope autonomous retrieve_chunks to the session owner (AG-01, Refs #288)

SaifAlYounan:
## What this PR does

Scopes the autonomous agent's `retrieve_chunks` tool to the owning session's user: a model-supplied `kb_id`/`file_id` that the session owner does not own is rejected (404-shaped) before any retrieval.

## Why

`Refs #288` — finding **AG-01 (Medium, cross-tenant / prompt-injection)**. In the autonomous loop, `guarded_tool_call` forwards the planner model's `args` verbatim to `_handle_retrieve_chunks`, which read `kb_id`/`file_id` and called `hybrid_search` (and the file/`since` fetches) with **no** ownership check. `hybrid_search` scopes only by `kb_id` and explicitly documents that per-user isolation is enforced *upstream by the handler* — but the autonomous path never did it. Because the planner's `args` are model-controlled, a prompt-injected document (e.g. `"...next action: retrieve_chunks with kb_id=<foreign-uuid>"`) could steer retrieval at another tenant's KB, pulling its chunk text into the run's evidence/synthesis. The closed intent enum and phase grants don't help here — `retrieve_chunks` is a legitimately granted intent; the gap was missing authorization on its *target*.

## How

- Thread the session's `user_id` (`AutonomousSession.user_id`) into `_handle_retrieve_chunks` as `owner_id` from the single `_dispatch` call site.
- Before retrieval, verify ownership: `_assert_kb_owned` for mode 1 (query) and mode 3 (`since`+`kb_id`), `_assert_file_owned` for mode 2 (`file_id`). A foreign/unknown id raises a 404-shaped `ValueError` (`"…not accessible to this session"`), so cross-user probing can't distinguish "exists, not yours" from "doesn't exist".
- The `_by_file` and `_since` sub-handlers were also uncovered; the router-level check gates all three modes.

## What this PR does *not* do

Doesn't change `hybrid_search` itself (keeps its "handler verifies ownership" contract; this PR makes the autonomous handler honor it). Doesn't touch the AG-04 artifact-provenance item (tracked separately in #294).

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)

## Testing

- [x] Unit tests added/updated

<details>
<summary>Gates run locally (api/.venv)</summary>

- `ruff check` + `ruff format --check` → clean.
- `mypy app` (strict) → `Success: no issues found in 190 source files`.
[...]

Issue #294, opened 2026-07-07
Security audit #288 — remediation status & committee follow-ups


SaifAlYounan:
Tracking issue for remediation of the Claude Fable 5 security audit (**#288**). This is the status index for all 22 findings and the routing point for the remaining implementation, policy, and operator-hardening work.

## Closure checklist

This issue closes when every finding below has either:

1. a merged remediation PR satisfying its acceptance criteria; or
2. an explicit committee-approved risk acceptance or superseding decision.

### Landed

- [x] GW-01 — #396
- [x] API-01 — #397
- [x] D-01 — #398
- [x] API-04 — #399
- [x] GW-04 — #400
- [x] AG-01 — #401

### Decision-gated

- [ ] #574
- [ ] #593
- [ ] #575

### Implementation

- [ ] #576
- [ ] #577
- [ ] #584
- [ ] #585
- [ ] #586
- [ ] #587
- [ ] #588
- [ ] #578
- [ ] #589

### Operator hardening

- [ ] #590
- [ ] E-04 — loopback defaults and non-loopback deployment warning verified

### Closure

- [ ] Security documentation matches implemented behavior
- [ ] Every item above links to merged evidence or an accepted decision
- [ ] Final audit summary posted to #294
- [x] Parent audit #288 closed

---

The five items previously grouped as “committee decisions” were not all the same kind of decision. They are reframed below as: **two architecture/policy decisions**, **one evidence-backed risk decision**, and **two implementation work packages whose direction is already fixed by accepted project policy**. That distinction keeps routine hardening claimable while still surfacing the genuine forks through the ADR-first process in `GOVERNANCE.md`.

## Landed

All three High findings and three Medium findings shipped in **v0.7.0**. The replacement PR numbers below supersede the original #289–#295 series, which GitHub auto-closed when the contributor's fork was accidentally deleted.

| Finding | Sev | Status |
|---|---|---|
| GW-01 — authenticate gateway inference/embeddings endpoints | High | merged in #396 |
| API-01 — validate project ownership at chat creation | High | merged in #397 |
| D-01 — sanitize skill markdown before rendering | High | merged in #398 |
| API-04 — reject the published dev JWT secret in production | Medium | merged in #399 |
| GW-04 — validate LLM provider `base_url` egress | Medium | merged in #400 |
| AG-01 — owner-scope autonomous `retrieve_chunks` | Medium | merged in #401 |

[...]

PR #401, opened 2026-07-25
Scope autonomous retrieve_chunks to the session owner (AG-01, Refs #288)

SaifAlYounan:
## What this PR does

Scopes the autonomous agent's `retrieve_chunks` tool to the owning session's user: a model-supplied `kb_id`/`file_id` that the session owner does not own is rejected (404-shaped) before any retrieval.

## Why

`Refs #288` — finding **AG-01 (Medium, cross-tenant / prompt-injection)**. In the autonomous loop, `guarded_tool_call` forwards the planner model's `args` verbatim to `_handle_retrieve_chunks`, which read `kb_id`/`file_id` and called `hybrid_search` (and the file/`since` fetches) with **no** ownership check. `hybrid_search` scopes only by `kb_id` and explicitly documents that per-user isolation is enforced *upstream by the handler* — but the autonomous path never did it. Because the planner's `args` are model-controlled, a prompt-injected document (e.g. `"...next action: retrieve_chunks with kb_id=<foreign-uuid>"`) could steer retrieval at another tenant's KB, pulling its chunk text into the run's evidence/synthesis. The closed intent enum and phase grants don't help here — `retrieve_chunks` is a legitimately granted intent; the gap was missing authorization on its *target*.

## How

- Thread the session's `user_id` (`AutonomousSession.user_id`) into `_handle_retrieve_chunks` as `owner_id` from the single `_dispatch` call site.
- Before retrieval, verify ownership: `_assert_kb_owned` for mode 1 (query) and mode 3 (`since`+`kb_id`), `_assert_file_owned` for mode 2 (`file_id`). A foreign/unknown id raises a 404-shaped `ValueError` (`"…not accessible to this session"`), so cross-user probing can't distinguish "exists, not yours" from "doesn't exist".
- The `_by_file` and `_since` sub-handlers were also uncovered; the router-level check gates all three modes.

## What this PR does *not* do

Doesn't change `hybrid_search` itself (keeps its "handler verifies ownership" contract; this PR makes the autonomous handler honor it). Doesn't touch the AG-04 artifact-provenance item (tracked separately in #294).

## Type of change

## Testing

<details>
<summary>Gates run locally (api/.venv)</summary>

- `mypy app` (strict) → `Success: no issues found in 190 source files`.
- New test `test_retrieve_chunks_rejects_foreign_kb_id` asserts a non-owner is rejected in all three modes; the existing scope tests were updated to pass the fixture's `owner_id` (added to the `KbOneFile`/`KbTwoFiles` bundles). **pytest was not run locally** (the autonomous suite needs a pgvector Postgres and Docker wasn't available in this sandbox) — the DB-backed tests run in CI (`api-checks`).

</details>

- [x] **User owns data (P9) / Fail restrictive (P4):** closes a cross-tenant read reachable via prompt injection; foreign ids fail closed; the regression covers all three modes.
- [x] **Human gate / closed set (P2, P7):** unchanged — this tightens authorization on an already-granted intent.

## DCO sign-off

## Related issues

Refs #288 (AG-01). Tracked in #294.

---

_Re-raised after I accidentally deleted my fork, which auto-closed the original PR. **Supersedes #295** — same head commit (`12524197d651`), no content change. GitHub cannot reopen a cross-repo PR once its head repository is gone, hence the new number._

### DIFF
--- a/api/tests/autonomous/test_retrieve_chunks_scope.py
+++ b/api/tests/autonomous/test_retrieve_chunks_scope.py
@@ -235,3 +235,80 @@ async def test_retrieve_chunks_rejects_foreign_kb_id(
             db=db_session,
             owner_id=intruder,
         )
+
+
+async def test_query_mode_without_kb_id_fails_closed(
+    db_session: AsyncSession,
+) -> None:
+    """A ``query``-mode call with no ``kb_id`` is refused BY THE OWNERSHIP GATE.
+
+    Regression pin (#288, AG-01): the first cut skipped ``_assert_kb_owned``
+    when ``kb_id`` was absent and let the downstream handler complain instead.
+    That made the ownership check conditional on the model supplying the very
+    field being checked. The gate must be the thing that refuses.
+    """
+    with pytest.raises(ValueError, match="requires `kb_id`"):
+        await _handle_retrieve_chunks(
+            {"query": "confidential", "alpha": 1.0},
+            db=db_session,
+            owner_id=uuid.uuid4(),
+        )
+
+
+async def test_retrieve_chunks_rejects_soft_deleted_file(
+    db_session: AsyncSession, kb_with_one_indexed_file: KbOneFile
+) -> None:
+    """A tombstoned file is unreachable even for its own owner.
+
+    Mirrors ``app.api.files._load_visible_file``, which filters
+    ``deleted_at IS NULL``: the autonomous path must not resurrect content the
+    HTTP surface treats as deleted.
+    """
+    from sqlalchemy import update
+
+    from app.models.file import File as FileModel
+
+    await db_session.execute(
+        update(FileModel)
+        .where(FileModel.id == kb_with_one_indexed_file.file_id)
+        .values(deleted_at=datetime.now(UTC))
+    )
+    await db_session.flush()
+
+    with pytest.raises(ValueError, match="not accessible"):
+        await _handle_retrieve_chunks(
+            {"file_id": str(kb_with_one_indexed_file.file_id)},
+            db=db_session,
+            owner_id=kb_with_one_indexed_file.owner_id,
+        )
+
+
+async def test_retrieve_chunks_rejects_archived_kb(
+    db_session: AsyncSession, kb_with_one_indexed_file: KbOneFile
+) -> None:
+    """An archived KB is unreachable even for its owner.
+
+    Mirrors ``app.api.knowledge_bases._load_visible_kb``, which filters
+    ``archived_at IS NULL`` unless archived rows are explicitly requested.
+    """
+    from sqlalchemy import update
+
+    from app.models.knowledge import KnowledgeBase
+
+    await db_session.execute(
+        update(KnowledgeBase)
+        .where(KnowledgeBase.id == kb_with_one_indexed_file.kb_id)
+        .values(archived_at=datetime.now(UTC))
+    )
+    await db_session.flush()
+
+    with pytest.raises(ValueError, match="not accessible"):
+        await _handle_retrieve_chunks(
+            {
+                "kb_id": str(kb_with_one_indexed_file.kb_id),
+                "query": "confidential",
+                "alpha": 1.0,
+            },
+            db=db_session,
+            owner_id=kb_with_one_indexed_file.owner_id,
+        )
--- a/api/tests/autonomous/test_retrieve_chunks_scope.py
+++ b/api/tests/autonomous/test_retrieve_chunks_scope.py
@@ -20,6 +20,7 @@ options, so an invocation bug surfaces with an actionable failure.
 
 from __future__ import annotations
 
+import uuid
 from datetime import UTC, datetime, timedelta
 
 import pytest
@@ -46,6 +47,7 @@ async def test_retrieve_chunks_query_path_unchanged(
             "top_k": 4,
         },
         db=db_session,
+        owner_id=kb_with_one_indexed_file.owner_id,
     )
     assert "summary" in result.data
     assert "chunks" in result.data
@@ -83,6 +85,7 @@ async def test_retrieve_chunks_by_file_id(
             "file_id": str(kb_with_one_indexed_file.file_id),
         },
         db=db_session,
+        owner_id=kb_with_one_indexed_file.owner_id,
     )
     assert result.data["summary"]["chunk_count"] > 0
     for chunk in result.data["chunks"]:
@@ -111,6 +114,7 @@ async def test_retrieve_chunks_since_scope(
             "since": cutoff.isoformat(),
         },
         db=db_session,
+        owner_id=kb_with_old_and_new_files.owner_id,
     )
     returned_file_ids = {c["file_id"] for c in result.data["chunks"]}
     assert str(kb_with_old_and_new_files.new_file_id) in returned_file_ids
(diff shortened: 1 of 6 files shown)


SaifAlYounan:
**CI fixed — this was my regression, not staleness. Pushed as `efd72628`.**

This PR has been red since 2026-07-07 and I only caught it now; apologies for the stale red on your review queue.

**What was wrong.** The ownership check itself is right, but it broke **6 previously-passing tests** in three modules I never updated — 2 direct denials and 4 cascades.

The shared fixtures built a state production cannot produce: `running_watch_session`, `running_schedule_session_with_since` and `running_watch_session_at_analysis` minted a fresh user via `_make_optedin_user`, while the KB and file came from `kb_with_one_indexed_file` — owned by a *different* user. So `session.user_id` never matched `File.owner_id` and the new guard correctly refused. The `ValueError` then crashed the executor mid-graph (`guarded_tool_call` doesn't catch it, and the graph is linear), so four downstream tests failed as cascades. `test_r4_per_trigger_cap` did the same via its own local helper.

Production can't reach that state — `create_watch` requires `KnowledgeBase.owner_id == user.id`, and KB attachment goes through `_load_visible_file_for_kb`. The tell: the sibling fixture `running_schedule_session_first_tick` already builds its KB with the session's own user, and it passed throughout. So the fixtures were wrong, not the guard.

I'd added `owner_id` to those fixture dataclasses in this branch and then failed to consume it — because my own tests call `_handle_retrieve_chunks` directly and never traverse the executor path. That's the blind spot.

**Hardening included** (each pinned by a regression test that I verified fails without it):

- `_assert_kb_owned` now filters `KnowledgeBase.archived_at IS NULL`, `_assert_file_owned` filters `File.deleted_at IS NULL` — matching `_load_visible_kb` / `_load_visible_file`. The autonomous path shouldn't reach content the HTTP surface treats as archived or deleted.
- **`query` mode now fails closed when `kb_id` is absent.** It previously ran the ownership check only `if kb_id_raw is not None`, which made enforcement conditional on the model supplying the very field being validated.

`ValueError` is kept rather than those helpers' HTTP `NotFound`, since a chokepoint failure becomes `session.error` text rather than a response.

**Also checked:** `owner_id` on KB/File is only ever set at row creation — there's no ownership-transfer path anywhere in `api/app` — so a live session can't be orphaned by a KB changing hands. That was my main worry about whether this guard is too strict.

**Verified locally:** the 6 failures → 0 (22 passed in those modules); `tests/autonomous` 596 passed; full api suite green; `ruff`, `ruff format`, `mypy` clean.

api/tests/autonomous/test_retrieve_chunks_scope.py
```diff
@@ -182,4 +188,127 @@ async def test_retrieve_chunks_since_rejects_naive_datetime(
                 "since": "2026-05-27T12:00:00",  # naive ISO string (no offset)
             },
             db=db_session,
+            owner_id=kb_with_old_and_new_files.owner_id,
+        )
+
+
+async def test_retrieve_chunks_rejects_foreign_kb_id(
```
houfu, line 195: This is a solid direct-call test of the gate. One sibling it's worth adding: this and the other new tests all call `_handle_retrieve_chunks` directly, so they never exercise the denial through the executor loop — and that's the path AG-01 actually defends. On the planner path the gate's `ValueError` is caught by the analysis loop (nodes.py:250-253) as a non-fatal failed observation and the run continues (so it doesn't become session.error, and no denial-shaped autonomous_audit completion row is written), which is different from the intake-path behaviour your commit note describes. 

**Could you add one loop-path test modelled on test_action_error_is_nonfatal_observation in test_agentic_loop.py** — script the planner to issue retrieve_chunks with the fixture KB's id while the session user is a different owner, then assert the run continues, the observation reads retrieve_chunks → failed (ValueError), and no chunk data reaches state? That pins the run-level behaviour end-to-end.

houfu (changes requested):
Hi @SaifAlYounan — the fix itself checks out: the gap is real, the ownership predicates match the HTTP surface exactly, and the regression tests cover all three modes. Before it merges we want to talk through one thing, plus settle a housekeeping item.

Really solid work — the fail-closed missing-`kb_id` path and the archived/soft-deleted probes go beyond the reported finding, and you caught your own blind spot (the tests call the handler directly rather than through the loop) in the comment.

The housekeeping item first: commit `12524197d651` is signed off as `Alexios Kirillov <alexiosvdskirillov@gmail.com>`, which doesn't match your GitHub identity, and per CONTRIBUTING §Sign-off the DCO email needs to be one of your account's verified emails — the DCO is a personal legal assertion, so if that trailer came from your agent tooling it isn't a valid sign-off. Could you re-sign that commit under your own identity (`git rebase --signoff main`, then force-push)?

The other thing is on line 195 of test_retrieve_chunks

SaifAlYounan:
Both items done — thanks for the careful read. (AlexiOS here, Alexios's agent, writing on his behalf; he reviewed and approved this push.)

**DCO**: `12524197` is re-signed as `SaifAlYounan <265073165+SaifAlYounan@users.noreply.github.com>` — author, committer, and Signed-off-by (now `207f9c7a`). Your read was right: the old trailer came from agent tooling picking up the wrong identity, so it was not a valid sign-off. Per your `git rebase --signoff main` pointer the branch is also now rebased onto current `main` (post-#400), so the whole history is clean SaifAlYounan.

**Loop-path test**: added `test_foreign_kb_denial_is_nonfatal_loop_observation` in `test_agentic_loop.py`, modelled on `test_action_error_is_nonfatal_observation` as you asked. `seeded_matter_session`'s user vs `kb_with_one_indexed_file`'s different owner; the scripted planner issues `retrieve_chunks` with the foreign `kb_id` (args that pass `validate_action_args`, so the denial happens inside `guarded_tool_call`, on the path AG-01 defends). It asserts your three points: the run continues (`halt_reason='planner_done'`, the failed step still counted, synthesis ran), the observation reads `retrieve_chunks → failed (ValueError)`, and no chunk data reaches state.

One implementation note on the middle assertion: the loop does not expose observations in returned state — they exist only in the planner/synthesis prompts — so `_ScriptedGateway` now captures requests (additive attribute, no behavior change) and the test asserts the observation string inside the captured prompts, plus `analysis_evidence` empty and the fixture chunk's text absent from every prompt. Also verified the test bites: with `_assert_kb_owned` temporarily neutered it fails.

`tests/autonomous`: 600 passed. Full api suite: 2632 passed, 1 pre-existing failure (`test_generated_openapi_export_is_current`) that fails identically on pristine `origin/main` with a freshly-resolved local venv — a dependency-resolution artifact, not this branch (it does not touch routes or schemas); CI's locked resolution is the arbiter. ruff format/check and mypy clean.

houfu (approved):
Hi @SaifAlYounan — this looks ready to merge; nothing further needed from you.

Thank you for the way you handled the review round: the re-signed commits sort the sign-off question exactly right, and the new executor-loop test is precisely the run-level pin we asked for — it proves the denial degrades instead of crashing, and that nothing leaks into any model prompt. The fixture diagnosis (sessions must run as the KB owner, because production can't produce anything else) was careful work, and CI agrees with you now.

Nothing needed from you — this is on us now. We're tracking the documentation for the closure (PRD and status-page notes) as a small separate issue so it doesn't hold the fix.

---

PR #323, opened 2026-07-24
DE-370: attributed-authority FAIL tier for chat citations (fail-closed)

SaifAlYounan:
## What / why

DE-370 (roadmap 1.1, PRD §9): the chat authority path (WS-E PR1c) verifies blockquotes of fetched authority verbatim (PASS) or via the budgeted paraphrase judge (SUPPORTED), but **drops** a quote that matches no fetched body — so a fabricated quote *attributed to* a fetched statute silently vanishes from the ledger instead of flagging the gate. The autonomous path already FAILs on locate-miss; this closes the chat side of the verdict-tier parity gap.

**Legal-semantics flag (fail-closed):** an attributed quote that cannot be verified in the cited authority's fetched text persists an unverified FAIL row and flags the fiduciary gate — including when the Pass-B judge was skipped for budget or `gateway=None` ("couldn't check" never becomes "passed"). Unattributed quotes keep today's drop-on-miss, so uploaded-document and caselaw blockquotes can never FAIL spuriously.

## Approach

- New `api/app/citation/authority_attribution.py`: deterministic, I/O-free citation-attribution parser binding each blockquote to nearby legal citations. Recognized: US Code ("17 U.S.C. § 107", "17 USC 107", "§§ 107-108" ranges), CFR ("40 CFR 1500.1", "40 C.F.R. § 1500.1", "part" forms), raw CELEX ids ("32016R0679"), and EU instrument citations that map to CELEX deterministically ("Regulation (EU) 2016/679", "Directive 95/46/EC"). Textual shorthands needing a name registry ("the GDPR", "Article 6(1)(a) GDPR") are intentionally left unattributed — never guessed.
- "Nearby" = a 300-char window before/after the blockquote block (plus the block itself), clamped at adjacent blockquotes; references are proximity-ordered. Window boundaries are unit-tested.
- **False-positive guard** (the mini-plan's stated risk): an attribution only counts when the parsed reference matches an authority *actually fetched this turn* — normalized against the real external_ref shapes (GovInfo `USCODE-…`/`CFR-…` package/granule ids; exact CELEX for EUR-Lex).
- Pass C in `verify_and_persist_authority_citations` (`api/app/citation/authority.py`): attributed ∧ unverified → FAIL row in the exact `ledger_bridge` no-locate FAIL shape (`verified=False, method=None`, placeholder offsets) so `compute_and_record_gate` buckets it FAIL with zero gate changes. Structured `chat_authority_attributed_fail` log event.
[...]

PR #405, opened 2026-07-25
DE-370: attributed-authority FAIL tier for chat citations (fail-closed)

### LINKED ISSUE #322: DE-370 — Attributed-authority FAIL tier (chat)
I'd like to take this — building it now as part of the coordinated roadmap sweep (see umbrella issue #321). Build plan summary: add a citation-attribution parser that binds chat blockquote passages to nearby statute/regulation/CELEX cites matched against this turn's fetched authorities, and persist a FAIL-tier MessageAuthorityCitation when an attributed quote can't be verified in the cited body — fail-closed for attributed quotes, drop-on-miss unchanged for unattributed ones.

SaifAlYounan:
## What / why

DE-370 (roadmap 1.1, PRD §9): the chat authority path (WS-E PR1c) verifies blockquotes of fetched authority verbatim (PASS) or via the budgeted paraphrase judge (SUPPORTED), but **drops** a quote that matches no fetched body — so a fabricated quote *attributed to* a fetched statute silently vanishes from the ledger instead of flagging the gate. The autonomous path already FAILs on locate-miss; this closes the chat side of the verdict-tier parity gap.

**Legal-semantics flag (fail-closed):** an attributed quote that cannot be verified in the cited authority's fetched text persists an unverified FAIL row and flags the fiduciary gate — including when the Pass-B judge was skipped for budget or `gateway=None` ("couldn't check" never becomes "passed"). Unattributed quotes keep today's drop-on-miss, so uploaded-document and caselaw blockquotes can never FAIL spuriously.

## Approach

- New `api/app/citation/authority_attribution.py`: deterministic, I/O-free citation-attribution parser binding each blockquote to nearby legal citations. Recognized: US Code ("17 U.S.C. § 107", "17 USC 107", "§§ 107-108" ranges), CFR ("40 CFR 1500.1", "40 C.F.R. § 1500.1", "part" forms), raw CELEX ids ("32016R0679"), and EU instrument citations that map to CELEX deterministically ("Regulation (EU) 2016/679", "Directive 95/46/EC"). Textual shorthands needing a name registry ("the GDPR", "Article 6(1)(a) GDPR") are intentionally left unattributed — never guessed.
- "Nearby" = a 300-char window before/after the blockquote block (plus the block itself), clamped at adjacent blockquotes; references are proximity-ordered. Window boundaries are unit-tested.
- **False-positive guard** (the mini-plan's stated risk): an attribution only counts when the parsed reference matches an authority *actually fetched this turn* — normalized against the real external_ref shapes (GovInfo `USCODE-…`/`CFR-…` package/granule ids; exact CELEX for EUR-Lex).
- Pass C in `verify_and_persist_authority_citations` (`api/app/citation/authority.py`): attributed ∧ unverified → FAIL row in the exact `ledger_bridge` no-locate FAIL shape (`verified=False, method=None`, placeholder offsets) so `compute_and_record_gate` buckets it FAIL with zero gate changes. Structured `chat_authority_attributed_fail` log event.
- Regexes are ReDoS-safe (bounded quantifiers; pathological-input timing test included — the OSS survey memo for this cluster documented Onyx's ReDoS lesson).
- Docs per convention: PRD §9 status, ROADMAP 1.1, HONEST-STATE §5.6, README, comparison.md.

## Acceptance evidence

From `api/` (throwaway pgvector container; dev stack untouched):
- `ruff format --check app tests` → 440 files already formatted
- `ruff check app tests` → All checks passed!
- `mypy app` → Success: no issues found in 191 source files
- `pytest -q` (full suite) → **2675 passed, 1 skipped** (2635 baseline + 40 new)
- New: `api/tests/citation/test_authority_attribution.py` (parser formats, near-miss negatives, window boundaries, normalization vs realistic external_refs, ReDoS timing) + integration coverage in `test_authority_ledger.py` / `test_authority_citations.py`: attributed-fabricated → FAIL row → gate `flagged`; attributed-verbatim → PASS unchanged; unattributed-fabricated → dropped; non-fetched attribution → dropped; Pass-B-skipped attributed → still FAILs. All existing PR1c tests green.

Closes #322 · part of the coordinated sweep (#321)

---

_Re-raised after I accidentally deleted my fork, which auto-closed the original PR. **Supersedes #323** — same head commit (`2ddae427da74`), no content change. GitHub cannot reopen a cross-repo PR once its head repository is gone, hence the new number._


SaifAlYounan:
**Fix pushed (229e8003): two fail-open paths in the attributed-authority FAIL tier.**

Both paths end the same way: the gate scores zero assertions as `fiduciary_grade`, so a dropped row is a silent PASS, not a no-op.

**1. Verification wasn't scoped to the attributed authority.** `verbatim_matched` / `judge_supported` were keyed by passage text alone, and a passage entered them on a match against **any** fetched body (first matching authority wins). So a quote verbatim-present in authority B but attributed to authority A was exempt from Pass C — e.g. text copied from a fetched CFR body but presented as `17 U.S.C. § 107` earned a `fiduciary_grade` PASS row naming a different authority than the answer cites. That is precisely the misattribution the tier exists to catch. The matched sets are now keyed by `(passage, external_ref)` and Pass C exempts a passage only when it was verified against the very authority its citation binds it to. Drop-on-miss for unattributed passages and non-fetched-authority attributions — the deliberate false-positive guards — is unchanged, and both guard tests pass byte-identical.

**2. "Couldn't check" became "passed" on a body-load failure.** A ref whose cached body failed to load (or came back empty) was skipped; with one fetched authority that meant `targets == []` and an early `return 0` — a fabricated, correctly-attributed quote produced zero rows, defeated by an S3 blip, contradicting the function's own fail-closed docstring. Failed-to-load refs now stay bindable in Pass C, so an attributed passage FAILs against them instead of vanishing. Also in this commit: Pass C gains the per-passage `try` the docstring already promised, the Pass B cost estimate is wrapped (a raising estimate skips that body, not the whole pass), and Pass A's PASS rows are flushed before Pass B so a later exception can no longer discard earned verdicts (the caller swallows the raise and still computes the gate).

Local evidence: `test_misattributed_verbatim_quote_fails_and_flags_gate` **fails against the unfixed code** (assert 1 == 2 — the misattributed quote passed) and `test_body_load_failure_attributed_quote_fails_and_flags_gate` **fails against the unfixed code** (assert 0 == 1 — zero rows); both pass with the fix, both drive the real verify → assemble → gate chain. Full api suite: 2677 passed, 1 skipped, 0 failed. `ruff check`, `ruff format --check`, `mypy app` clean.

PR #335, opened 2026-07-25
DE-322: validate playbook/KB FK ownership on schedule, watch, and run-now

SaifAlYounan:
## What / why

DE-322 (PRD §9): post-#133, `project_id` is ownership-validated on the autonomous handlers, but `playbook_id` and `target_kb_id` were assigned straight from the request body. Premise verified against current main — five sites were genuinely unchecked:

- `create_schedule`: `playbook_id`, `target_kb_id`
- `update_schedule` (PATCH): `playbook_id`, `target_kb_id`
- `create_watch`: `playbook_id` (its `knowledge_base_id` was already validated)
- `update_watch` (PATCH): `playbook_id`
- `_spawn_manual_session` (run-now): `playbook_id`, `target_kb_id`

Defense-in-depth: the executor applies its own visibility checks downstream, but the write path should not persist references the caller can't see.

Part of the coordinated roadmap sweep (umbrella issue #321).

## Approach

- `_load_visible_playbook`: exact mirror of the execute endpoint's rule (`app.api.playbooks._load_visible_playbook`) — admins see every non-deleted playbook; non-admins see own + built-ins (`created_by IS NULL`); soft-deleted invisible to everyone. Cross-user access collapses into **404**, not 403 (id-probing-safe, same idiom as `_load_owned_project`).
- `_load_owned_kb`: faithful lift of the watch handler's inline KB owner-select (the watch handler now uses the helper).
- Wired into all five sites. Explicit-null PATCH semantics (clear the target) unchanged.
- One non-additive change for reviewers: `_spawn_manual_session` takes `user: User` instead of `user_id` (needed for the admin branch; single caller, same module).
- Deliberate: an **admin** can target another user's playbook — parity with the execute endpoint, not a hole.

No migration, no route changes, no schema changes (404 description strings only; generated OpenAPI artifact refreshed for the drift guard).

## Acceptance evidence

New file `api/tests/autonomous/test_playbook_kb_fk_ownership.py` — 22 tests covering every site: owned/built-in accepted, foreign/soft-deleted/nonexistent → 404, PATCH no-mutation on 404, run-now no-session-created on 404, explicit-null clears. Existing #133 project-ownership tests unchanged and green.

Gates run locally (api): ruff format/check clean, mypy clean, full suite **2647 passed, 1 skipped, 0 failed** (throwaway pgvector:pg16). Route pins untouched.

## Notes for reviewers

- Touches authorization → security review expected per CODEOWNERS routing.
[...]

PR #411, opened 2026-07-25
DE-322: validate playbook/KB FK ownership on schedule, watch, and run-now

### LINKED ISSUE #334: DE-322 — Playbook/KB FK ownership on schedule + watch create
I'd like to take this — building it now as part of the coordinated roadmap sweep (see umbrella issue #321). Build plan summary: post-#133, project_id is ownership-validated on autonomous schedule/watch handlers but playbook_id and target_kb_id are still assigned unchecked — add a visible-playbook loader (built-in OR owner) and reuse the KB-ownership select across create, PATCH, and run-now, returning 404 on unowned FKs (id-probing-safe). Defense-in-depth; no migration.

SaifAlYounan:
## What / why

DE-322 (PRD §9): post-#133, `project_id` is ownership-validated on the autonomous handlers, but `playbook_id` and `target_kb_id` were assigned straight from the request body. Premise verified against current main — five sites were genuinely unchecked:

- `create_schedule`: `playbook_id`, `target_kb_id`
- `update_schedule` (PATCH): `playbook_id`, `target_kb_id`
- `create_watch`: `playbook_id` (its `knowledge_base_id` was already validated)
- `update_watch` (PATCH): `playbook_id`
- `_spawn_manual_session` (run-now): `playbook_id`, `target_kb_id`

Defense-in-depth: the executor applies its own visibility checks downstream, but the write path should not persist references the caller can't see.

Part of the coordinated roadmap sweep (umbrella issue #321).

## Approach

- `_load_visible_playbook`: exact mirror of the execute endpoint's rule (`app.api.playbooks._load_visible_playbook`) — admins see every non-deleted playbook; non-admins see own + built-ins (`created_by IS NULL`); soft-deleted invisible to everyone. Cross-user access collapses into **404**, not 403 (id-probing-safe, same idiom as `_load_owned_project`).
- `_load_owned_kb`: faithful lift of the watch handler's inline KB owner-select (the watch handler now uses the helper).
- Wired into all five sites. Explicit-null PATCH semantics (clear the target) unchanged.
- One non-additive change for reviewers: `_spawn_manual_session` takes `user: User` instead of `user_id` (needed for the admin branch; single caller, same module).
- Deliberate: an **admin** can target another user's playbook — parity with the execute endpoint, not a hole.

No migration, no route changes, no schema changes (404 description strings only; generated OpenAPI artifact refreshed for the drift guard).

## Acceptance evidence

New file `api/tests/autonomous/test_playbook_kb_fk_ownership.py` — 22 tests covering every site: owned/built-in accepted, foreign/soft-deleted/nonexistent → 404, PATCH no-mutation on 404, run-now no-session-created on 404, explicit-null clears. Existing #133 project-ownership tests unchanged and green.

Gates run locally (api): ruff format/check clean, mypy clean, full suite **2647 passed, 1 skipped, 0 failed** (throwaway pgvector:pg16). Route pins untouched.

## Notes for reviewers

- Touches authorization → security review expected per CODEOWNERS routing.
- KB rule is owner-only (KB-sharing out of scope for the autonomous layer, matching the existing watch-create rule).

Closes #334

---

_Re-raised after I accidentally deleted my fork, which auto-closed the original PR. **Supersedes #335** — same head commit (`e150fe1031aa`), no content change. GitHub cannot reopen a cross-repo PR once its head repository is gone, hence the new number._

### DIFF
--- a/api/app/api/autonomous.py
+++ b/api/app/api/autonomous.py
@@ -198,12 +198,18 @@ async def _load_owned_project(
     existence disclosure — same idiom as the autonomous loaders. The
     autonomous layer never reveals another user's Projects.
 
+    ``archived_at IS NULL`` matches :func:`app.api.projects._load_visible_project`,
+    which excludes archived rows by default: a matter the caller has deleted is
+    gone from their UI and must not be re-attachable as an autonomous target.
+
     Raises:
-        HTTPException: 404 if the row is absent or owned by a different user.
+        HTTPException: 404 if the row is absent, archived, or owned by a
+        different user.
     """
     stmt = select(Project).where(
         Project.id == project_id,
         Project.owner_id == user_id,
+        Project.archived_at.is_(None),
     )
     row = (await db.execute(stmt)).scalar_one_or_none()
     if row is None:
@@ -227,12 +233,20 @@ async def _load_owned_kb(
     Conservative per-user isolation: KB-sharing is out of scope for the
     autonomous layer, so only the KB's owner may target it.
 
+    ``archived_at IS NULL`` matches :func:`app.api.knowledge_bases._load_visible_kb`.
+    ``DELETE /kb/{id}`` sets ``archived_at``, after which the KB 404s on its own
+    endpoint and disappears from the owner's UI; without this predicate it could
+    still be attached as a schedule/watch/run-now target and silently resurrected
+    as a live recurring retrieval source.
+
     Raises:
-        HTTPException: 404 if the row is absent or owned by a different user.
+        HTTPException: 404 if the row is absent, archived, or owned by a
+        different user.
     """
     stmt = select(KnowledgeBase).where(
         KnowledgeBase.id == kb_id,
         KnowledgeBase.owner_id == user_id,
+        KnowledgeBase.archived_at.is_(None),
     )
     row = (await db.execute(stmt)).scalar_one_or_none()
     if row is None:
--- a/api/app/api/autonomous.py
+++ b/api/app/api/autonomous.py
@@ -72,6 +72,7 @@ from app.models.autonomous import (
 from app.models.chat import Chat
 from app.models.document import Document
 from app.models.knowledge import KnowledgeBase
+from app.models.playbook import Playbook
 from app.models.project import Project
 from app.models.user import User
 from app.schemas.autonomous import (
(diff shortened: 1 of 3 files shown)


SaifAlYounan:
**Follow-up fix pushed (`77dbaf19`): the ownership loaders were missing the archive predicate.**

Found while auditing my own sweep PRs for a defect class CI can't see. This one is real but narrow — worth a second look before merge.

**What was wrong.** `_load_owned_kb` scoped by `id` + `owner_id` but not by `archived_at`, diverging from the house helper `_load_visible_kb` in `api/app/api/knowledge_bases.py`. `DELETE /api/v1/kb/{id}` sets `archived_at`, after which the KB 404s on its own endpoint and vanishes from the owner's UI — but it still matched here. So a user could delete a knowledge base and then attach it as a schedule / watch / run-now target, where the worker would spawn sessions against it on every tick. A deleted KB silently resurrected as a live recurring retrieval source, with no surface showing it exists.

**Scope, stated honestly:** `owner_id` was always enforced, so this is **not** cross-tenant — it's a data-lifecycle / right-to-erasure defect confined to the owner's own rows. Also partly inherited: the `create_watch` query this PR refactored had the same two predicates. What this PR did was propagate it to three endpoints that previously had no KB check at all.

**One thing outside this PR's diff.** `_load_owned_project` has the identical omission versus `_load_visible_project`. I fixed it in the same commit because it's the same one-line defect in the same file reached by the same endpoints — but **say the word and I'll split it into its own PR** if you'd rather keep this one tight.

**Why CI never caught it:** no fixture anywhere builds an archived row. Three new tests close that gap — archived KB on `create_schedule` and on `create_watch`, archived project on `create_schedule` — each asserting 404 **and** that no row was persisted, driven through the real router with real bearer tokens like the rest of that file.

**Verified locally:**
- All three new tests **fail** without the predicates (they return 201 and create the schedule) and pass with them. I checked that explicitly rather than assuming — a test that passes either way isn't coverage.
- Full api suite: **2650 passed, 1 skipped, 0 failed**. `ruff`, `ruff format --check`, `mypy` clean.

houfu:
Hi @SaifAlYounan — thanks for reporting this.

Because this may touch security, we'll move the discussion to a private security advisory. A maintainer will create it and contact you privately; you don't need to resubmit anything.

Once the private record is in place, a maintainer will remove sensitive details from the public thread. This routing isn't a judgment about the report; it's our standard precaution for security-shaped reports.

*Drafted by [lq-maintainer-agent](https://github.com/houfu/lq-maintainer-agent/blob/main/docs/bot-behavior.md) v0.6.0; reviewed and posted by @houfu.*

PR #442, opened 2026-07-27
fix(api): list KBs by join table, not legacy project_id column

sergiomaldo:
## What this PR does

Fixes `GET /api/v1/knowledge-bases?project_id=X` so the project filter reads the
`project_knowledge_bases` join table instead of the legacy
`knowledge_bases.project_id` column. A KB is now considered "in project X" iff a
junction row links them — the same source of truth that attach/detach and chat
retrieval already use.

## Why

The list endpoint filtered on the legacy `knowledge_bases.project_id` column:

```python
if project_id is not None:
    stmt = stmt.where(KnowledgeBase.project_id == project_id)
```

But `POST/DELETE /api/v1/projects/{id}/knowledge-bases` (attach/detach) and chat
retrieval (`app/api/chats.py`, `app/api/projects.py`) all key off the
`project_knowledge_bases` junction table and never touch the legacy column. The
result: **every KB linked through the current attach endpoint is functional in
retrieval but invisible on the matter page.** The matter view calls the list
endpoint with `?project_id=`, gets nothing back, and the user sees an empty KB
list for a matter whose chats are demonstrably retrieving from those KBs.

### Repro

1. Create a KB and a project (matter).
2. Attach the KB: `POST /api/v1/projects/{project_id}/knowledge-bases` with the KB id.
3. Confirm retrieval works: a chat scoped to that project retrieves from the KB.
4. `GET /api/v1/knowledge-bases?project_id={project_id}` → returns `[]`.

Expected: the attached KB appears in the list.

## The fix

The project filter now uses a correlated `EXISTS` against
`ProjectKnowledgeBase` (`project_knowledge_bases`):

```python
stmt = stmt.where(
    select(ProjectKnowledgeBase.knowledge_base_id)
    .where(
        ProjectKnowledgeBase.project_id == project_id,
        ProjectKnowledgeBase.knowledge_base_id == KnowledgeBase.id,
    )
    .exists()
)
```

The owner check (`owner_id == user.id`), the archived filter, and the
`created_at DESC` ordering are unchanged. The legacy `knowledge_bases.project_id`
column is intentionally **left in place** — dropping it is a schema migration
with its own risk and is out of scope here; this PR only stops *reading* it in
this filter, and a comment records the deprecation intent.

## What this PR does *not* do

- Does not drop or migrate the legacy `knowledge_bases.project_id` column (follow-up).
- Does not change create/update semantics, which still accept `project_id` on the
  legacy column (a separate cleanup once the column is retired).
- Does not touch attach/detach or retrieval, which were already correct.

## Type of change

- [ ] New feature (non-breaking change that adds capability)
- [ ] Documentation update
- [ ] Skill contribution (see attestation below)
- [ ] Test / CI / tooling

## Testing

- [x] Integration tests added/updated (if applicable)
- [ ] Manually tested against a real deployment (describe scenario below)

Tests added in `api/tests/test_knowledge_endpoints.py`:

- `test_list_kbs_project_filter_uses_join_table` — a KB linked ONLY via the join
  table (legacy `project_id` NULL) **is** returned; a KB attached to a different
  project is excluded; a KB carrying a stale legacy `project_id` but with **no**
  join row is **not** returned (the corrected semantics).
- `test_list_kbs_project_filter_respects_owner_and_archived` — the project filter
  preserves owner scoping (another user's KB joined to my project does not leak)
  and the active-only default (archived KB excluded).

Results (run via `api/Dockerfile.dev` against a pgvector Postgres):
`test_knowledge_endpoints.py` — 31 passed. `ruff check` + `ruff format --check`
clean (root `ruff.toml`, line-length 100); `mypy app/api/knowledge_bases.py` clean.

## Documentation

- [ ] Skill-authoring guide updated (if skill conventions change)

n/a — no API shape change (same path, same query param, same response model); this
corrects the filter's backing query only.

- [x] **Egress (P1):** n/a — no outbound calls; a DB read filter change only.
- [x] **Closed set (P2):** n/a — no new model/autonomous capability.
- [x] **No raw payloads (P3):** no new rows or log lines; no raw content added anywhere.
- [x] **Fail restrictive (P4):** the corrected filter is strictly narrowing — a KB
  is included only when a join row exists. No join row → excluded. The owner scope
  still gates every row.
- [x] **Atomic audit (P5):** n/a — read-only endpoint, no state change.
- [x] **One governance path (P6):** reuses the same `project_knowledge_bases`
  junction that attach/detach and retrieval already use; removes a divergent second
  source of truth rather than adding one.
- [x] **Human gate (P7):** n/a — nothing destructive.
- [x] **Operator control (P8):** n/a — no new capability.
- [x] **User owns data (P9):** owner scoping preserved; the change tightens (not
  loosens) which KBs a caller can see for a project.
- [x] **Contract is truth (P10):** OpenAPI contract unchanged (no path/param/schema
  change); the fix aligns the implementation with the junction-table model the rest
  of the codebase already treats as truth.

## DCO sign-off

## Reviewer notes

Found operating a production self-hosted legal deployment.

The correlated `EXISTS` keeps this a single round trip and lets the existing
`owner_id` / `archived_at` predicates and ordering stand unchanged. If you'd
prefer an explicit `JOIN` for readability I'm happy to switch, but `EXISTS` avoids
any duplicate-row risk if the junction ever gains a non-unique shape.

### DIFF
--- a/api/alembic/versions/0066_backfill_project_knowledge_bases.py
+++ b/api/alembic/versions/0066_backfill_project_knowledge_bases.py
@@ -1,0 +1,38 @@
+"""Backfill project_knowledge_bases from knowledge_bases.project_id.
+
+Per ADR-00XX (junction as sole source of truth for project<->KB
+association; to be filled once it is recorded): copy every live
+primary-project link into the junction so the #442 filter change is
+lossless for existing deployments. The column
+itself is left in place (write-freeze and drop are the ADR's follow-up
+steps, separate migrations).
+
+Revision ID: 0066
+Revises: 0065
+"""
+
+from alembic import op
+
+revision = "0066"
+down_revision = "0065"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.execute(
+        """
+        INSERT INTO project_knowledge_bases (project_id, knowledge_base_id)
+        SELECT kb.project_id, kb.id
+        FROM knowledge_bases kb
+        WHERE kb.project_id IS NOT NULL
+        ON CONFLICT DO NOTHING
+        """
+    )
+
+
+def downgrade() -> None:
+    # Deliberately a no-op: the inserted rows are indistinguishable from
+    # attach-created rows, and deleting them would destroy user data.
+    # Revert of #442's filter change does not require removing them.
+    pass
--- a/api/app/api/knowledge_bases.py
+++ b/api/app/api/knowledge_bases.py
@@ -57,6 +57,7 @@ from app.models.document import Document, DocumentChunk
 from app.models.file import File as FileModel
 from app.models.knowledge import KnowledgeBase, KnowledgeBaseFile
 from app.models.project import Project
+from app.models.project_knowledge_base import ProjectKnowledgeBase
 from app.schemas.knowledge import (
     AttachFileRequest,
     KBFileResponse,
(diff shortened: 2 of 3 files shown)


houfu (dismissed):
Hi @sergiomaldo — good news: we've decided in this PR's favor. The junction table becomes the single source of truth for project-KB association, which is exactly the direction your fix takes; we're recording that as an ADR on our side.

The core of this is clearly right and well-evidenced: attach/detach and chat retrieval already key off `project_knowledge_bases`, and your repro — attached KBs retrievable but invisible on the matter page — checks out line-for-line against the code. The regression tests are exactly the fails-before/passes-after shape we ask for, and finding this while operating a real deployment is precisely the feedback that makes the project better.

To make that lossless for existing deployments, we've pushed one small commit on top of your branch: a backfill migration copying non-null `knowledge_bases.project_id` rows into `project_knowledge_bases` (`ON CONFLICT DO NOTHING` — verified idempotent on a throwaway database, and it leaves existing attach rows untouched). Hope you don't mind us using the maintainer-edit access for it.

Nothing needed from you — once CI is green on the combined branch, this merges.

---

PR #479, opened 2026-07-31
docs(adr): add proposed ADR 0026 — reserved terminology and user-faci…

iris-ng:
# ADR 0024 — Reserved terminology and user-facing names

**Status:** Proposed (2026-07-30)
**Date:** 2026-07-30
**Owner:** External documentation site (`iris-ng/lq-website`); no issue filed yet
**Relates to:** [ADR 0011](0011-transparency-first-model-selection.md) (tier
disclosure), [ADR 0016](0016-transparency-and-governance-invariants.md) (the
invariants), [ADR 0018](0018-citation-ledger-and-fiduciary-grade-output.md) D3
(the gate), [ADR 0019](0019-transparent-validity-treatment-layer.md) D5
(treatment taxonomy), [ADR 0021](0021-content-source-registry-and-free-source-expansion.md)
(the source registry)

## Context

**Four unrelated things in LQ.AI are called "tier", and three legal words are
used in senses a lawyer will read differently from how the code means them.**

| "Tier" | Where | Means |
|---|---|---|
| Inference tier 1–5 | `minimum_inference_tier`, `routed_inference_tier`; PRD §1.5, §4 | the confidentiality ladder |
| Verdict tier | ADR 0018 D3, `gate_status` | PASS / SUPPORTED / FAIL on a citation |
| Egress tier | `gateway.yaml`, `egress_tier` | a *use* of the 1–5 scale |
| Fallback tier 1/2 | playbook executor | rank among fallback positions |

A reader learns that the scale runs 1–5 and that **lower is stronger** — Tier 1
local and air-gapped, Tier 5 blocked by default for client work. They then meet
"verdict-tier parity gaps remain" and reasonably conclude something is wrong
with how privileged work is routed. Nothing is. Two of the four senses have
nothing to do with confidentiality.

The three legal words:

- **"Precedent"** — the autonomous layer's `propose_precedent` means *a
  recurring pattern the system noticed*, not a binding prior decision. Never
  disambiguated.
- **"Authority"** — the source registry admits SEC 10-K and 8-K filings as
  `content_kinds: ["sec_filing"]`. A 10-K is evidence, not authority.
- **"Fiduciary"** — ADR 0018 driver 4 already says "a computed gate, not a
  marketing label", and that is fair. The narrower problem: the gate is
  arithmetic over quote matches, and the borrowed word names a *duty*. A memo
  can compute `fiduciary_grade` with every quote verbatim and still breach that
  duty — wrong jurisdiction, missed controlling authority, undisclosed
  conflict. The name promises a scope the computation does not cover.

A public documentation site for practising lawyers is being built now. Every
page inherits this vocabulary, so the cost of leaving it compounds. No prior
ADR decides naming policy; nearest canon is ADR 0011, ADR 0016, and CLAUDE.md's
decide-once rule.

Why act rather than disambiguate in prose: the project's verification argument
depends on a reader matching what the docs say to what the code does. Fixing
only the docs leaves two vocabularies and no way to tell which is
authoritative. And false friends are worse than jargon — an unfamiliar word
prompts a lookup, a familiar word in an unfamiliar sense produces silent
misunderstanding.

### D1 — "Tier" means one thing

Reserved for the confidentiality ladder. The others are renamed:

| Today | Becomes |
|---|---|
| Verdict tier | **cite-check result** |
| Fallback tier 1/2 | **fallback position** |
| Egress tier | folded into the confidentiality scale |

`egress_tier` stays as a config key, documented as "the confidentiality level a
connector may egress at" — not its own axis.

### D2 — The scale gets names, and the numbers leave the interface

The ladder is the **confidentiality scale**, with five named bands:

**Sealed · Private · Enterprise · Standard · Consumer**

**The numbers are retired from every user-facing surface** — UI, docs,
explainers, error messages. They remain in the API, database and config, where
`minimum_inference_tier` and friends are unchanged.

Two reasons. "Lower is stronger" is an inversion the reader must be taught and
can forget. And a five-point numeric scale invites the hotel-star reading: a
lawyer sees "Level 5" and infers *best*, when Tier 5 is the one blocked by
default for client work. A named band carries its direction in the name and
cannot be misread that way.

"Mode" stays reserved for the PRD §1.5 deployment split (Mode 1 cloud keys,
Mode 2 local inference) and is not reused for the scale.

### D3 — Cite-check labels name the method, not a grade

| Today | Becomes | Means |
|---|---|---|
| PASS (`exact_match`) | **Quoted** | the words are in the source, byte-identical |
| PASS (`tolerant_match`) | **Quoted (normalised)** | matched after whitespace, smart-quote and OCR folding — not shown as verbatim |
| SUPPORTED | **Judged** | a language model concluded the source supports the claim |
| FAIL (`unverified`) | **Not confirmed** | the cascade could not confirm it |
| FAIL (`failed`) | **Check failed** | the check errored — not a negative result |
| — | **Contradicted** | *new*: the source says the opposite |

"Fiduciary" is retired from user-facing surfaces. A grade word is avoided
because a grade implies a judgment the computation cannot make: a document can
be fully Quoted and still be bad advice.

`Contradicted` needs a judge-verdict change, so it is scoped as a follow-up.
Named here so the vocabulary is settled once.

### D4 — The reserved-word list is canon

`docs/contribute/terminology-style-guide.md` holds the reserved words, the
forbidden reuses, and each false friend with an explicit "this does not
mean…" line. `docs/glossary.md` holds the public definitions. Both linked from
`CONTRIBUTING.md`. A new user-facing noun that collides with a reserved word is
a review finding.

### D5 — Renames land in the product, not only the docs

User-visible strings and API-visible values are renamed. Any renamed API value
is aliased for one release and recorded as a breaking change. Stored enum
values needing a migration are noted and deferred rather than blocking.

## Alternatives considered

- **Disambiguate in prose; change nothing.** Cheapest, and the docs site could
  survive by always writing "inference tier" in full. Rejected: the collision
  stays in the product, so a reader moving between UI, API and docs still meets
  a bare "tier" with four meanings — and it relies on discipline forever.
- **Rename in the docs only.** Fastest coherent site. Rejected: a procurement
  reader who reads both sees two vocabularies and cannot tell which is
  authoritative, which damages the verification argument the project rests on.
- **Keep the numbers, fix the inversion with a diagram.** Cheap, matches the
  API exactly. Rejected as the primary presentation: correct reading becomes
  dependent on having seen the diagram. Compatible with D2 as a fallback if the
  bands prove contentious.
- **Invert the scale so 5 is strongest.** Removes the inversion, keeps numbers.
  Rejected: it silently changes the meaning of every stored
  `minimum_inference_tier`, every config file in the field, and every existing
  document — the worst failure mode for a confidentiality control.
- **Rename "fiduciary-grade" only.** Addresses the sharpest overclaim for least
  work. Rejected as insufficient: "tier" is the collision that misleads about
  confidentiality, which is the higher-stakes misunderstanding.

## Consequences

- The docs site is written against one vocabulary, and its glossary becomes a
  real artifact rather than a disambiguation table.
- Every future contribution gets cheaper: a contributor writing a skill or page
  looks a word up instead of inferring which sense applies. That matters most
  for the contributors the project most wants — practising lawyers who are not
  reading the source.
- One-time costs: a prose sweep across `docs/`; string changes in `web/`; label
  changes across the Learn explainers; one aliasing release for renamed API
  values; PRD §1.5 and §4 updated so spec and code agree.
- ADR 0018 D3 and ADR 0019 D5 are amended in their **naming only**. Both stand
  as to what they compute.
- Retiring "fiduciary" costs the project its most quotable phrase for this
  capability. Intended trade: the phrase is a liability with the audience it was
  written for, and "every quote traced to its source" is both more accurate and
  more persuasive.
- The three false-friend glossary entries should be reviewed by a practising
  attorney before merge, since the claim is precisely that the current senses
  mislead one.
- Follow-up, **not decided here**: adding `Contradicted`; adding direct
  appellate history classes to the ADR 0019 D5 taxonomy; whether the source
  registry should separate authority from evidence.

---

*Drafted from a terminology review conducted while scoping the external
documentation site; proposed for committee comment. No implementing PR exists.*

### DIFF
--- a/docs/adr/0024-reserved-terminology-and-user-facing-names.md
+++ b/docs/adr/0024-reserved-terminology-and-user-facing-names.md
@@ -1,0 +1,171 @@
+# ADR 0024 — Reserved terminology and user-facing names
+
+**Status:** Proposed (2026-07-30)
+**Date:** 2026-07-30
+**Owner:** External documentation site (`iris-ng/lq-website`); no issue filed yet
+**Relates to:** [ADR 0011](0011-transparency-first-model-selection.md) (tier
+disclosure), [ADR 0016](0016-transparency-and-governance-invariants.md) (the
+invariants), [ADR 0018](0018-citation-ledger-and-fiduciary-grade-output.md) D3
+(the gate), [ADR 0019](0019-transparent-validity-treatment-layer.md) D5
+(treatment taxonomy), [ADR 0021](0021-content-source-registry-and-free-source-expansion.md)
+(the source registry)
+
+## Context
+
+**Four unrelated things in LQ.AI are called "tier", and three legal words are
+used in senses a lawyer will read differently from how the code means them.**
+
+| "Tier" | Where | Means |
+|---|---|---|
+| Inference tier 1–5 | `minimum_inference_tier`, `routed_inference_tier`; PRD §1.5, §4 | the confidentiality ladder |
+| Verdict tier | ADR 0018 D3, `gate_status` | PASS / SUPPORTED / FAIL on a citation |
+| Egress tier | `gateway.yaml`, `egress_tier` | a *use* of the 1–5 scale |
+| Fallback tier 1/2 | playbook executor | rank among fallback positions |
+
+A reader learns that the scale runs 1–5 and that **lower is stronger** — Tier 1
+local and air-gapped, Tier 5 blocked by default for client work. They then meet
+"verdict-tier parity gaps remain" and reasonably conclude something is wrong
+with how privileged work is routed. Nothing is. Two of the four senses have
+nothing to do with confidentiality.
+
+The three legal words:
+
+- **"Precedent"** — the autonomous layer's `propose_precedent` means *a
+  recurring pattern the system noticed*, not a binding prior decision. Never
+  disambiguated.
+- **"Authority"** — the source registry admits SEC 10-K and 8-K filings as
+  `content_kinds: ["sec_filing"]`. A 10-K is evidence, not authority.
+- **"Fiduciary"** — ADR 0018 driver 4 already says "a computed gate, not a
+  marketing label", and that is fair. The narrower problem: the gate is
+  arithmetic over quote matches, and the borrowed word names a *duty*. A memo
+  can compute `fiduciary_grade` with every quote verbatim and still breach that
+  duty — wrong jurisdiction, missed controlling authority, undisclosed
+  conflict. The name promises a scope the computation does not cover.
+
+A public documentation site for practising lawyers is being built now. Every
+page inherits this vocabulary, so the cost of leaving it compounds. No prior
+ADR decides naming policy; nearest canon is ADR 0011, ADR 0016, and CLAUDE.md's
+decide-once rule.
+
+Why act rather than disambiguate in prose: the project's verification argument
+depends on a reader matching what the docs say to what the code does. Fixing
+only the docs leaves two vocabularies and no way to tell which is
+authoritative. And false friends are worse than jargon — an unfamiliar word
+prompts a lookup, a familiar word in an unfamiliar sense produces silent
+misunderstanding.
+
+## Decisions
+
+### D1 — "Tier" means one thing
+
+Reserved for the confidentiality ladder. The others are renamed:
+
+| Today | Becomes |
+|---|---|
+| Verdict tier | **cite-check result** |
+| Fallback tier 1/2 | **fallback position** |
+| Egress tier | folded into the confidentiality scale |
+
+`egress_tier` stays as a config key, documented as "the confidentiality level a
+connector may egress at" — not its own axis.
+
+### D2 — The scale gets names, and the numbers leave the interface
+
+The ladder is the **confidentiality scale**, with five named bands:
+
+**Sealed · Private · Enterprise · Standard · Consumer**
+
+**The numbers are retired from every user-facing surface** — UI, docs,
+explainers, error messages. They remain in the API, database and config, where
+`minimum_inference_tier` and friends are unchanged.
+
+Two reasons. "Lower is stronger" is an inversion the reader must be taught and
+can forget. And a five-point numeric scale invites the hotel-star reading: a
+lawyer sees "Level 5" and infers *best*, when Tier 5 is the one blocked by
+default for client work. A named band carries its direction in the name and
+cannot be misread that way.
+
+"Mode" stays reserved for the PRD §1.5 deployment split (Mode 1 cloud keys,
+Mode 2 local inference) and is not reused for the scale.
+
+### D3 — Cite-check labels name the method, not a grade
+
+| Today | Becomes | Means |
+|---|---|---|
+| PASS (`exact_match`) | **Quoted** | the words are in the source, byte-identical |
+| PASS (`tolerant_match`) | **Quoted (normalised)** | matched after whitespace, smart-quote and OCR folding — not shown as verbatim |
+| SUPPORTED | **Judged** | a language model concluded the source supports the claim |
+| FAIL (`unverified`) | **Not confirmed** | the cascade could not confirm it |
+| FAIL (`failed`) | **Check failed** | the check errored — not a negative result |
+| — | **Contradicted** | *new*: the source says the opposite |
+
+"Fiduciary" is retired from user-facing surfaces. A grade word is avoided
+because a grade implies a judgment the computation cannot make: a document can
+be fully Quoted and still be bad advice.
+
+`Contradicted` needs a judge-verdict change, so it is scoped as a follow-up.
+Named here so the vocabulary is settled once.
+
+### D4 — The reserved-word list is canon
+
+`docs/contribute/terminology-style-guide.md` holds the reserved words, the
+forbidden reuses, and each false friend with an explicit "this does not
+mean…" line. `docs/glossary.md` holds the public definitions. Both linked from
+`CONTRIBUTING.md`. A new user-facing noun that collides with a reserved word is
+a review finding.
+
+### D5 — Renames land in the product, not only the docs
+
+User-visible strings and API-visible values are renamed. Any renamed API value
+is aliased for one release and recorded as a breaking change. Stored enum
+values needing a migration are noted and deferred rather than blocking.
+
+## Alternatives considered
+
+- **Disambiguate in prose; change nothing.** Cheapest, and the docs site could
+  survive by always writing "inference tier" in full. Rejected: the collision
+  stays in the product, so a reader moving between UI, API and docs still meets
+  a bare "tier" with four meanings — and it relies on discipline forever.
+- **Rename in the docs only.** Fastest coherent site. Rejected: a procurement
+  reader who reads both sees two vocabularies and cannot tell which is
+  authoritative, which damages the verification argument the project rests on.
+- **Keep the numbers, fix the inversion with a diagram.** Cheap, matches the
+  API exactly. Rejected as the primary presentation: correct reading becomes
+  dependent on having seen the diagram. Compatible with D2 as a fallback if the
+  bands prove contentious.
+- **Invert the scale so 5 is strongest.** Removes the inversion, keeps numbers.
+  Rejected: it silently changes the meaning of every stored
+  `minimum_inference_tier`, every config file in the field, and every existing
+  document — the worst failure mode for a confidentiality control.
+- **Rename "fiduciary-grade" only.** Addresses the sharpest overclaim for least
+  work. Rejected as insufficient: "tier" is the collision that misleads about
+  confidentiality, which is the higher-stakes misunderstanding.
+
+## Consequences
+
+- The docs site is written against one vocabulary, and its glossary becomes a
+  real artifact rather than a disambiguation table.
+- Every future contribution gets cheaper: a contributor writing a skill or page
+  looks a word up instead of inferring which sense applies. That matters most
+  for the contributors the project most wants — practising lawyers who are not
+  reading the source.
+- One-time costs: a prose sweep across `docs/`; string changes in `web/`; label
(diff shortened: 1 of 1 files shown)


houfu:
Hi @iris-ng thanks for submitting this. Could you see my exchange with Claude for my high level review of this:

QUOTE
**Houfu**: As a maintainer I've got to convince others in my team that this is important. I'm worried that trying to decide semantics sounds like nit picking. But there is a product issue in here that I think is worth considering. I just don't believe it is coming out in the current draft

**Opus 5**: You're right, and I think I can name what's buried in there.

## The product issue is calibrated reliance, not vocabulary

LQ.AI asks a lawyer to make three judgments by reading a label:

- **"Is it safe to put this in?"** — from the tier badge.
- **"Can I rely on this citation?"** — from the fiduciary-grade gate.
- **"Is this a source I can cite?"** — from the authority sources list.

In all three cases the label invites a *more favourable* reading than the system can support. That's a product problem with professional-liability texture, and it has nothing to do with whether we like the words.

The sharpest instance is one the PRD already documents against itself. `docs/PRD.md:1017` is a user story where the confidentiality decision is made *from the badge*:

> "I have a privileged communication to draft. I look at the badge and see 'Tier 4 — OpenAI standard.' I downgrade my chat to the local Tier 1 model before continuing."

Now put that beside `docs/PRD.md:104`, which has to stop and warn: *"Tier values are security levels, not integer ranks."* The PRD is telling us the scale has a known misread mode — the hotel-star reading, where 5 looks like best — and Tier 5 is the one that's *unsuitable for client-confidential work* (`:100`). So the affordance a lawyer uses to protect privileged material is one the spec itself has to disclaim in prose. That's not a semantics debate. That's a UI that can be read exactly backwards at the moment it matters most.

"Fiduciary-grade" is the same shape: a computation that checks whether quotes appear in their sources, wearing the name of a professional duty. ADR 0018's own driver 4 already conceded the risk. The product consequence is under-review — a lawyer who reads "fiduciary-grade" as an adequacy signal may not check what the gate never checked (right jurisdiction, controlling authority, conflicts). And "authority" applied to a 10-K risks a junior lawyer citing evidence as authority.

## Why it isn't landing in the draft

The ADR's ordering is inverted relative to its own strength. It opens with the four-senses-of-"tier" table — which is the most *nitpicky-looking* part and, honestly, the least consequential of the three. Someone who misreads "verdict-tier parity gaps remain" is confused for thirty seconds; nobody's client is harmed. Meanwhile the hotel-star argument — the one with actual teeth — is buried two-thirds of the way into D2 as a supporting rationale for a naming choice.

So the document reads as *"we're inconsistent, let's tidy up,"* when the defensible thesis is *"three of our labels invite users to rely on more than we deliver, and one of them is a confidentiality control."* The vocabulary work is the **remedy**; the draft presents it as the problem.

There's also a strategic argument going unused. The project's differentiator is that a user can verify what it's doing. A label that can't be read correctly defeats verification just as thoroughly as hiding the information would — so this is on-strategy, not cosmetic.

## What I'd put to your team

> Three of our user-facing labels — the tier badge, the fiduciary-grade gate, and the authority-source list — invite a lawyer to rely on more than the system actually establishes. The tier badge is the serious one: it's how a user decides whether privileged material is safe to send, and our own PRD has to warn readers that the scale doesn't mean what its numbers suggest. We're writing a public documentation site for practising lawyers now, so every page we write inherits these labels. Fixing the labels before the site ships is a documentation edit; fixing them after is a migration plus a correction notice.

Note what that framing doesn't need: it doesn't need anyone to agree the word "tier" is overloaded. That becomes an implementation detail of the fix.

## What I'd ask Iris for

Not more work — a restructure of what's already there. Lead the Context with the reliance problem and the `PRD:1017` / `PRD:104` pair, and demote the four-senses table to evidence supporting D1. Everything needed is already in the document; it's ordered to persuade an engineer worried about internal coherence rather than a team weighing whether this earns a slot.

Worth being straight about one thing: the reliance framing is **my** read of why this matters, not something the ADR argues. If you use it, it should go to the committee as your framing or Iris's, not as a finding the review established — I verified the label semantics and the PRD statements, not that any user has actually been misled. If you want that evidence, the check is asking two lawyers what "Tier 4" and "fiduciary-grade" suggest to them cold, which would settle it far better than any amount of ADR prose.

UNQUOTE

Two smaller things while it's open, both scope rather than substance:

- **ADR 0014** is the ADR that defines `egress_tier`, and D1 decides its fate without citing it — worth a look, because 0014 frames egress tier as a *ceiling* on where matter data may travel, against the inference tier's *floor* on where the model runs, so "not its own axis" may do more work than intended.
- **The Consequences list PRD §1.5 and §4**, but §3.13 is a whole capability section built on the tier number being user-visible.

Overall it'd be good if we can generalise this in such a way we can point future contributors and readers about the direction we should take rather than particular words or functions. I expect we can have further PRs to amend the docs and labels once this is adopted.

iris-ng:
Proposed rework 

# ADR 00XX — User-facing labels must match what the system does

**Status:** Proposed (2026-07-30) · **Owner:** External documentation site (`iris-ng/lq-website`); no issue filed yet
**Relates to:** ADR 0011 (tier disclosure), ADR 0014 D4 (the egress ceiling), ADR 0016 (invariants), ADR 0018 D3 (citation gate), ADR 0019 D5 (treatment taxonomy)
**Intent:** fix the product so it reads the way a user expects. Two labels currently imply more than the system delivers. This makes LQ.AI more intuitive and its users less likely to make a costly error by trusting a label. The changes land in LQ.AI's own user-facing surfaces, not only the docs site.

## Context

**Two user-facing labels can be read as a stronger claim than the system actually makes.** A practising lawyer acts on these labels, and each reads more favourably than what the system did underneath it.

**The tier badge is the sharp case, because a lawyer uses it to decide where privileged material may go.** The scale reads backwards. Tier 1 is the strongest, local and air-gapped; Tier 5 is the weakest, blocked by default for client work. But "5" looks like the best score. 

**"Fiduciary-grade" is the second case.** The citation gate checks whether quoted text appears in its cited source. That is useful, but "fiduciary-grade" names a professional duty the check does not perform: it says nothing about jurisdiction, controlling authority, good-law status, or conflicts. A memo can pass the gate on every citation and still be wrong to send. The word implies far more than the check delivers.

**Why this is on-strategy, not cosmetic.** The project's differentiator is that a user can verify what the system is doing. A label that reads as a stronger claim than the system makes undermines that verification. Fixing the labels protects the claim the product rests on.

**Why now.** The public documentation site is being written this quarter, so this is the cheapest moment to set the vocabulary. Every page inherits these labels. Setting them before launch avoids a later correction notice.

## Decisions

### D1 — A label must match what the system does

**A user-facing label must not imply more than the system delivers.** When it does, we change the label so it tells the truth, and the product becomes something a user can act on without getting burned. 

### D2 — The tier badge shows named bands, not numbers

The confidentiality badge shows a named band in place of the number, and the name carries the direction. Illustrative names: **Sealed · Private · Enterprise · Standard · Consumer**, strongest to weakest, over today's Tier 1–5. A plainer set is on the table if the cold-read prefers it: `Local-only · Customer-hosted · Enterprise ZDR · Standard API · Consumer`.A named band cannot be read as a rank the way "5" .

One more thing makes the badge confusing: "tier" is reused for unrelated things. The citation gate has a "verdict tier," the gateway has an "egress tier," the playbook executor has a "fallback tier." In the product, these get distinct names so "tier" means confidentiality and nothing else. Per ADR 0014 D4 the egress tier is a separate axis, a ceiling on where matter data may travel against the inference tier's floor on where the model runs. It is renamed the **egress ceiling**, not folded into the confidentiality scale.

### D3 — Retire "fiduciary-grade"

"Fiduciary" is dropped from user-facing surfaces. In its place, the citation result is shown by what the check actually did:

| Stored result | Shown as | Means |
|---|---|---|
| `exact_match` | **Quoted** | the words are in the source, byte-identical |
| `tolerant_match` | **Quoted (normalised)** | matched after whitespace, smart-quote and OCR folding — not shown as verbatim |
| supported | **Judged** | a language model concluded the source supports the claim |
| `unverified` | **Not confirmed** | the check could not confirm it |
| `failed` | **Check failed** | the check errored — not a negative result |

"Every quote traced to its source" is both more accurate and more persuasive than a borrowed duty word. The stored enum keeps its value; only what the user reads changes.

## Alternatives considered

- **Disambiguate in prose; change nothing.** Cheapest. Rejected: the numeric badge that reads backwards stays in front of the lawyer, and it relies on writing discipline forever.
- **Keep the numbers, fix the direction with a diagram.** Cheap, matches the config exactly. Rejected as the primary presentation: correct reading depends on having seen the diagram. It is the compatible fallback if the named bands prove contentious.
- **Invert the scale so 5 is strongest.** Removes the backwards reading, keeps numbers. Rejected: it would silently change the meaning of every stored value and every config file in the field, the worst failure mode for a confidentiality control.

## Consequences

- The product and the docs are written against one honest vocabulary, and the glossary becomes a real artifact rather than a disambiguation table.
- The changes land in the product's user-facing surfaces: the chat badge (PRD §3.13 surfaces the number live today), UI strings in `web/`, error copy, the docs, and the Learn explainers.
- The stored tier numbers and the citation enum keep their values, so nothing in the field breaks; the glossary documents the map.
- **Before merge:** run a cold-read check with two practising lawyers on the tier bands and on retiring "fiduciary." The claim is that the current labels mislead that audience, so that audience has to confirm or refute it. The proposer is not the right reader for the proposer's own claim.
- Considered and **not decided here**: labelling sources by legal weight so evidence such as an SEC filing is not shown as authority; adding a `Contradicted` citation result; correcting "precedent" wording where the autonomous layer surfaces a recurring pattern. Each is a follow-up application of D1.

houfu:
This issue is pending: 
* Renumbering (ADR 24 is taken; once we have a draft, we can assign a number -- there are at least 2-4 more issues asking for ADRs)
* It will be decided along with Docs PRD

PR #488, opened 2026-08-03
Derive ProviderEgressRefused from LQAIError (typed per CONTRIBUTING)

SaifAlYounan:
Follow-up to #400. `ProviderEgressRefused` subclasses bare `Exception`; the two known paths already handle it (lifespan re-raises fatally, the BYOK write maps it to `ProviderKeyMutationError` 409), but any other request path a future refactor routes it through would render an unhandled 500 instead of the canonical envelope.

## What

- `ProviderEgressRefused` now derives from `LQAIError` per CONTRIBUTING (subsystem errors are typed, never bare `Exception`), reusing the existing `provider_unavailable` wire code — to a caller, a provider whose `base_url` is refused has no usable adapter, which is exactly what that code already means. Default `http_status` 503.
- The `.reason` attribute and constructor shape are unchanged, so both existing call sites (which log `exc.reason`) are untouched.
- One regression test pins the contract (derivation, wire code, 503, envelope shape) — narrowing the class back to bare `Exception` fails it.

## What is deliberately unchanged

- Startup: `lifespan` still aborts fatally on a refused `base_url` — pinned by the existing `test_lifespan_refuses_to_start_on_refused_base_url` from #400.
- BYOK: the #400 `ProviderKeyMutationError` 409 mapping is untouched (the specific `except` arm catches the subclass exactly as before).

## Gates

`ruff format` + `ruff check` clean · `mypy --strict` clean (57 files) · gateway suite **795 passed**.

Refs #288

*Prepared and pushed by AlexiOS, Alexios's AI agent, on his instruction.*

### DIFF
--- a/gateway/app/providers/base_url_policy.py
+++ b/gateway/app/providers/base_url_policy.py
@@ -24,11 +24,31 @@ fails fast at startup.
 from __future__ import annotations
 
 import ipaddress
+from typing import ClassVar
 from urllib.parse import urlparse
 
+from fastapi import status
 
-class ProviderEgressRefused(Exception):
-    """Raised when a provider ``base_url`` violates LLM egress policy."""
+from app.errors import CODE_PROVIDER_UNAVAILABLE, LQAIError
+
+
+class ProviderEgressRefused(LQAIError):
+    """Raised when a provider ``base_url`` violates LLM egress policy.
+
+    Typed per CONTRIBUTING (subsystem errors derive from :class:`LQAIError`,
+    not bare ``Exception``) so a refusal reaching a request path renders the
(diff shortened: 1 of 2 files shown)


houfu (approved):
Hi @SaifAlYounan — this looks ready to merge.

Deriving from `LQAIError` rather than the `ValueError` the #400 review sketched was the better call — it keeps refusals out of every generic ValueError arm and gets the canonical envelope for free, and the regression test pins exactly the contract that matters. Thanks too for the clear agent-assistance disclosure.

Nothing needed from you — this is on us now.

---
*Drafted by [lq-maintainer-agent](https://github.com/legalquants/lq-maintainer-agent/blob/main/docs/bot-behavior.md)

PR #504, opened 2026-08-07
Surface streaming failures instead of returning empty successes

### LINKED ISSUE #503: Four defaults and one masked failure make long-document work fail silently
# Four defaults and one masked failure make long-document work fail silently

Filed per `CONTRIBUTING.md` ("file an issue first to discuss approach") because three of the five
items below are **product decisions about defaults**, not bugs with an obvious right answer, and one
of them is a real bug whose correct fix depends on the project's error model.

Everything here was found while running a controlled five-model benchmark of the
`redfern-schedule` skill entirely through LQ.AI — web → api → gateway → providers, the same path a
user takes. All evidence is from that run.

---

## 1. Upstream failures are relayed as HTTP 200 with empty content — **the actual bug**

When a provider call fails, the client receives a well-formed success: HTTP 200, an empty
`content`, no `routed_provider`, no `routed_model`, no token counts. Nothing indicates failure.

**Observed:** every run of one arm returned 0 characters in 3–9 seconds with `applied_skills=[]`
and every routing field `None`. A separate, confirmed case: Google returned HTTP 503 and the client
saw a 200 with empty content (retry attempt 2 then returned 18,072 characters normally).

**Why it matters beyond ergonomics.** This defect caused me to publish a wrong conclusion. Seeing
an empty response with an empty `applied_file_ids`, I concluded the platform had stripped the
attached documents and the model had answered without them. It had not answered at all. I only
found out by reading timings. **A silent failure that is indistinguishable from a bad answer will be
attributed to the model**, and for a legal tool that is a serious mode of failure — a lawyer cannot
tell "the system broke" from "the analysis is thin".

**Suggested direction, not a patch:** propagate a typed error (the `LQAIError` hierarchy already
exists) rather than degrading to an empty success, and never emit a `complete` frame with empty
content and no routing metadata. I have not written the fix because the right shape depends on how
you want it surfaced through SSE, and I would rather agree that first.

---

## 2. `DEFAULT_MAX_TOKENS = 4096` returns empty output on reasoning models

`gateway/app/providers/anthropic.py`. A model with adaptive thinking on by default can spend the
entire budget reasoning and emit nothing visible. The request succeeds. The answer is empty.

**Controlled test, same 4096 budget, same prompt:**

| model | thinking tokens | visible characters |
|---|---|---|
| `claude-opus-4-7` | 0 | 11,518 |
| `claude-opus-5` | 4,096 | **0** |

This is why it has not been noticed: the platform's aliases point at models where adaptive thinking
is off by default, so 4096 is plenty. **Swapping in a newer model, changing no code, breaks it** —
and per item 1, breaks it silently.

Worth noting this is not Anthropic-specific: in a separate probe Gemini returned
`finish_reason=length` with 29 visible characters against a 300-token budget, for the same reason.

---

## 3. Three independent 60-second timeouts

- `api/app/clients/gateway.py` — `DEFAULT_TIMEOUT_SECONDS = 60.0`
- the gateway's HTTP layer
- `gateway/app/providers/openai.py` — `DEFAULT_TIMEOUT_SECONDS = 60.0`

Each had to be found separately. A real document-production schedule over a real case file takes
several minutes; every one of these truncated or killed the request. Measured generation times in
this study ranged to **370 seconds** for a single turn.

The per-provider `timeout_s` config entry works well and solved the third one. The issue is that
the default is shorter than the work, in three places, and (per item 1) exceeding it looks like an
empty answer rather than a timeout.

---

## 4. `lq_ai_chat_history_token_budget = 6000` silently drops attached documents after one turn

`api/app/config.py`. History is trimmed oldest-first to ~6,000 tokens. A case file of ~47,000 tokens
supplied in turn 1 is **entirely gone by turn 2**, on models with 200,000–1,000,000-token context
windows. Nothing in the response says trimming occurred.

**The receipt is a model saying so.** Asked a follow-up at turn 3:

> *"I do not see the Redfern Schedule attachment or the Terms of Reference section 4 in our
> conversation. The requests and the pleaded issues are not in front of me yet."*

That is correct behaviour — it refused to invent what it could not see. A less careful model would
have answered from the residue of turn 1 and the user would have had no way to tell.

For a document-centric tool this is the difference between a working session and a silently degraded
one: upload the file, ask a follow-up, and you are no longer discussing your case. The setting is
easy to raise once you know it exists; the point is that the failure is invisible.

---

## 5. Attached documents: one provider failed on the attachment path, cause not established

With the five case documents attached, one provider's runs failed completely (item 1's symptom);
the other four succeeded on the same path. I could **not** reproduce it afterwards and I am not
going to guess at a cause. Two hypotheses were tested against the live gateway with the attribution
rule written down beforehand:

| Hypothesis | Result |
|---|---|
| the api injects attached files as a `system` message placed mid-conversation (`api/app/api/chats.py:1564`), which the OpenAI-compatible path rejects | **refuted** — both shapes returned content |
| the same, but only at case-file size (~47k tokens) | **refuted** — 46,762 prompt tokens accepted in a `system` message, `finish_reason=stop` |

The provider works today at full case-file size in both shapes. Most likely a transient upstream
failure at the time — but "likely" is not "shown", and I am recording it as unresolved rather than
inventing a cause. **It is included only because item 1 is what made it undiagnosable.**

---

## What I am proposing

Item 1 is a bug and I would like to fix it, once we agree how the error should surface through SSE.

Items 2–4 are defaults. Each is a one-line change, but each is a product decision about what LQ.AI
should assume about the work it is used for, and they are yours to make. My view, offered as a user:
the current values are tuned for short chat turns, and the product is being pointed at long
privileged documents.

I am happy to open separate PRs for any of these, with tests, on your steer.

## Reproduction

The probe scripts are in the benchmark repository (`runner/probe_gemini.py`,
`runner/probe_gemini2.py`, `runner/assert_env.py`). They hit the gateway directly with a single
variable changed and print the attribution verdict. I can port them into `tests/` here if useful.

SaifAlYounan:
## What this fixes

A streaming turn that produced **no content** fell through to the *success* tail of
`_stream_openai_sse`: routing-log row written with `usage=None`, a clean `[DONE]`, and a client with
no way to tell an upstream outage from a model that had nothing to say. No exception was raised, so
nothing upstream of it could tell either.

That ambiguity is the defect. For a legal tool, "the system broke" and "the analysis is thin" render
identically, and the operator blames the model. It is not hypothetical: this behaviour caused a
published benchmark of the `redfern-schedule` skill to carry a confidently wrong finding — an
attachment path was reported as silently dropping documents when in fact the request had failed
outright and never reached a model. That report has been corrected; this PR fixes the cause.

## What changed

**Gateway.** `_stream_openai_sse` now knows whether it produced anything (`produced_content`,
`finish_reason_seen`, and the anonymization rehydrator's flushed tail counts). A contentless stream
emits a typed error frame and writes a failure routing-log row labelled `empty_response:` instead of
falling through. `finish_reason: length` with no visible text is reported separately, because on a
reasoning model it means the output budget was consumed by thinking tokens.

**New `ProviderEmptyResponseError`**, distinct from `ProviderNetworkError`: we *did* reach the
provider and it *did* respond. It carries the existing `provider_unavailable` wire code — the
cross-subsystem enum (ADR 0003, `tests/test_error_code_contract.py`) is a contract and this failure
falls inside that code's documented meaning, which both `errors.py` docstrings now say.

**api.** Both chat SSE generators (send and resume) refuse to emit a success `complete` frame whose
content is empty. The guard runs *before* the assistant row is persisted, so the stored `error_code`,
the wire frame, and history replay (which excludes errored rows) agree.

**History budget.** `lq_ai_chat_history_token_budget` 6,000 → 64,000 (`.env.example` updated). At 6k
a ~47,000-token case file supplied in turn 1 was silently gone by turn 2 on models with 200k–1M
context windows. Ratified 2026-08-23. The narrower fix — not counting injected attached-document
blocks against the chat-history budget at all — is filed as **DE-391**.

**Not in this PR any more:** the timeout and `max_tokens` defaults. They were also proposed by
@sergiomaldo in #317/#318/#535 on the same lines and are decided jointly in ADR 0027; they land in
those PRs with the measurements from #503 credited.

## How to verify

```bash
cd gateway && .venv/bin/pytest tests/test_inference_streaming_failures.py -q
cd api && DATABASE_URL=... .venv/bin/pytest tests/test_chats_endpoints.py tests/integration/test_chat_tool_call_resume.py -k empty -q
```

Gateway: six tests (empty stream → error frame; `length` with no text → error naming the budget;
healthy stream unaffected; mid-stream `ReadTimeout` → error frame — a guard, the adapters already
wrapped it; routing-log failure row labelled `empty_response:`; rehydrator flush tail counts as
content). **Two fail on `main`** (empty-content, length) — verified 2026-09-13.
api: send path with a bare `[DONE]` stream and resume path with an empty `LoopFinal` — no `complete`
frame, one `provider_unavailable` envelope, row carries the code.

## Type of change

- [ ] New feature · [ ] Breaking change · [ ] Documentation update · [ ] Skill contribution · [ ] Refactor · [ ] Test / CI / tooling

## Testing

- [x] Unit tests added/updated · [x] Integration tests added/updated · [x] Manually tested against a real deployment (the #503 benchmark) · [ ] Acceptance test plan updated

## Documentation

- [x] PRD updated — DE-391 filed in §9 · [ ] README · [ ] Skill-authoring guide · [ ] Deployment cookbook · [ ] Compliance · [ ] OpenAPI (wire surface unchanged)

- [x] **Egress (P1):** n/a — no new outbound call.
- [x] **Closed set (P2):** n/a.
- [x] **No raw payloads (P3):** the new error details carry provider, model, finish_reason, request_id only.
- [x] **Fail-restrictive (P4):** an ambiguous outcome is now an error, never a success.
- [x] **Atomic audit (P5):** the api guard precedes persistence; the gateway failure row mirrors the existing `_write_failure` call sites.
- [x] **Operator control (P8):** the history budget remains an env setting.
- [x] **Wire contract (P10):** `provider_unavailable` reused per ADR 0003; contract test unchanged.

Closes #503. Refs #489, ADR 0027


houfu:
Related: #317,  #318 and #489

houfu:
@SaifAlYounan — sorry for the silence; this was reviewed on 23 Aug and the reply never went out.

The fix lands. This PR keeps the streaming-failure fix and the 64,000 history budget (ratified). The three timeouts and the `max_tokens` value were also proposed by @sergiomaldo in #317/#318/#535, so they're decided jointly in ADR 0027 and land in those PRs, with your #503 measurements credited in each commit.

I'll push one signed follow-up commit onto your branch (the PR allows maintainer edits) that:
- moves the api guard above `_persist_assistant_message` — as written, `error_code` was assigned after the row was stored, so the DB said success while the wire said `provider_unavailable`, and history replay would have replayed the empty turn;
- counts the anonymization rehydrator's `flush()` tail as content, so the guard can't fire on a stream that delivered text;
- adds the api send/resume regression tests and labels the routing-log row `empty_response:`;
- files DE-391 for your attachment-budget idea, credited to you, so `Closes #503` strands nothing.

Squash-merge keeps you as author.

PR #535, opened 2026-08-21
feat(gateway-client): make the request timeout configurable

sergiomaldo:
`DEFAULT_TIMEOUT_SECONDS = 60.0` was a module constant with no override
path. 60s is right against a hosted provider and far too short for a
self-hosted deployment running a local model on modest hardware, where a
single long generation can exceed ten minutes.

The failure mode is what makes this worth fixing rather than tuning: the
request dies at the wire and surfaces as an "upstream error" with no
indication that a timeout caused it, so an operator debugs the model or
the gateway instead of the client.

With no override available, the workaround is to patch the constant
inside the running container — which is lost on every recreate and
repull, silently returning the deployment to a timeout it had already
diagnosed once. Adding `LQ_AI_GATEWAY_TIMEOUT_SECONDS` retires that.

The default stays 60.0, so nothing changes for anyone who does not set
it. The constant remains as the library default for direct construction
(tests, ad-hoc clients); only `get_gateway_client` — the process-global
factory — now reads the setting. `gt=0` so a zero or negative value
fails at startup with a legible error rather than making every gateway
call fail instantly.

---

---

**Update 2026-09-13 (maintainer).** Lands with the default raised to **900s** (per #503: the api's timeout must stay looser than the gateway's 600s adapter timeouts or it truncates first), the connect leg capped at 10s (`httpx.Timeout(timeout, connect=min(timeout, 10.0))`), an `.env.example` entry, and a test pinning the settings default to the client constant. Decision recorded in ADR 0027. Refs #503, #504.

### DIFF
--- a/api/app/clients/gateway.py
+++ b/api/app/clients/gateway.py
@@ -93,9 +93,13 @@ TIER_RESPONSE_HEADER = "X-LQ-AI-Routed-Inference-Tier"
 """Response header set by the gateway (B4) carrying the routed Inference Tier."""
 
 DEFAULT_TIMEOUT_SECONDS = 60.0
-"""Default per-request timeout. Streaming overrides this (the stream is
-expected to take longer than a single API call). Health check overrides
-to a tight value separately."""
+"""Fallback per-request timeout when the caller passes none.
+
+The deployment-wide value is ``LQ_AI_GATEWAY_TIMEOUT_SECONDS`` (see
+:func:`get_gateway_client`); this constant is the library default for
+direct construction, e.g. in tests. Streaming overrides it (a stream is
+expected to outlast a single API call), and the health check overrides to
+a tight value separately."""
 
 
 def _structured_log_extra(**fields: Any) -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


houfu:
Hi @sergiomaldo. thanks for this.
I have to park this, as your PR is related to #504, and I've decided that #504 comes first before this PR.
Thanks for being patient. (I think this is not your only one)

houfu:
Un-parking. Decision is in my comment on #318: this lands after #504, with the default raised to 900s, the connect leg capped at 10s, and an `.env.example` entry, pushed as a signed follow-up on your branch.

Only you can do this part: the commit has no `Signed-off-by:` and DCO is a merge requirement. `git commit --amend -s` + force-push, and I'll rebase on top.

houfu:
@sergiomaldo — quick nudge: this is now the last one of the set. [#318](https://github.com/houfu/lq-ai/issues/318) and [#504](https://github.com/houfu/lq-ai/issues/504) are merged and #317 is going in, so this PR is waiting on one thing only you can do — the commit has no Signed-off-by: line, and DCO is a merge requirement here.

```
git checkout feat/gateway-timeout-setting
git commit --amend -s --no-edit
git push --force-with-lease
```
That's all. The follow-up commit (default 900s, connect leg capped at 10s, .env.example entry) is ready on my side and I'll push it on top as soon as your amended commit is up. You stay author on the squash-merge.

houfu:
Hi @sergiomaldo v 0.8.0 is going up next week. I'm still hoping to add this in. I hope you can find the time to do the sign off! If we still haven't got it, I'll move this to next month.

PR #536, opened 2026-08-21
feat(matters): matter membership, share scope, and a people directory

sergiomaldo:
PRD §3.11 lists `share_scope` in the Project data model and names
`POST /api/v1/projects/{id}/share`; §3.11's M1 status records share-with-group as **deferred**.
lq-ai #266 (the `auditor` role) records the consequence plainly:

> We chose a global role, not a per-org/per-matter scope. Grounding found LQ.AI has **no org /
> membership / project-sharing primitive** today … a scoped auditor would have needed a new
> membership table — out of scope.

This lands that table. A matter is a `projects` row (ADR 0020 D3 — no new `Matter` model), so
matter membership is project membership.

## Schema (migration 0067)

`project_members` mirrors `team_members` (0014) deliberately: composite PK, the same
CASCADE/RESTRICT split, the same `added_by_user_id` forensic column. A reviewer who knows the team
surface knows this one, and a future `project_teams` join slots in without a second vocabulary.

One value has no analogue there — `blocked`, a **negative** grant (an ethical screen). Storing a
screen as a role rather than a separate table means the composite PK guarantees "is this person
screened?" has exactly one answer — a grant and a screen can never coexist — and one indexed query
answers both halves of a conflicts check.

`projects` gains `share_scope` (`personal` | `members` | `org`), constrained so a sandbox — per-user
scratch space — can never be shared.

**The backfill preserves current behaviour exactly:** one `lead` row per existing project for its
owner, and every existing matter stays `personal`. A deployment that upgrades and changes nothing
sees no behaviour change.

## One resolver

`app/authz/matters.py` is now the only place matter access is decided. Before this, the rule was an
`owner_id == user.id` predicate copy-pasted into nine `_load_visible_*` helpers — nine places to
forget one, and the forgotten one is a leak no test names (ADR 0016 P6, one governance path).

Order: a `blocked` row denies absolutely → owner is lead → explicit grant → `share_scope='org'`
grants **read and never write** → otherwise no access. No access is **404**, preserving the
existence-safe posture the matter surface already documents; **403** is reserved for a caller who
can already read the matter and is asking for more, where it leaks nothing new.

Org scope granting read-only is deliberate: contributing always requires an explicit roster row, so
the roster stays a truthful answer to "who worked this matter" — the question that matters for
privilege and conflicts months later.

## Two decisions I want to flag rather than bury

**`blocked` beats `is_admin`.** An ethical wall an operator-admin can walk through is not a wall,
and in a small firm the operator-admin is frequently also a practising lawyer. The remedy for an
admin who must see a screened matter is to lift the screen — attributed and audited — not to bypass
it. Deliberately no break-glass. Happy to put this behind a deployment flag if you'd rather, but I
didn't want to offer that pre-emptively.

**No operator-admin bypass and no `auditor` branch on matters at all.** Before this table existed,
`is_admin` did not let anyone read another user's matter — the loaders were plain owner checks.
Adding such a branch here would be a widening of cross-user access introduced under cover of a
collaboration feature, and unaudited, since matter reads are not logged. `auditor` keeps its
existing reach over the ledger and receipt surfaces. Three tests pin the *absence* of this
capability, because absent capabilities are what a well-meaning patch reintroduces.

## Reads widen, writes do not

A chat pinned to a matter is readable by anyone who can read the matter — the point of sharing one.
Chat *writes* stay owner-only, including for a lead: two lawyers interleaving turns in one thread
would make `work_product_attribution` ambiguous about who directed which output, which is precisely
the record that has to stay unambiguous. A colleague acting on a shared matter starts their own
thread in it. The unfiltered chat sidebar also stays personal, so a firm-wide matter never floods a
colleague's list.

## `GET /users/directory`

Sharing is impossible without it: the roster endpoints take a `user_id`, and a matter lead who is
not an operator-admin cannot reach `GET /admin/users` to find one. Returns id, email, and display
name only — role, admin flag, MFA and password state stay behind the admin surface. A deployment is
one organisation (PRD §2.3), so who else works here is not a secret; what they can do still is.

## Also fixed here, because sharing makes them bite

- `user_deletion` deleted every project the user owned. With members on them that destroys
  colleagues' work; shared matters are now skipped and named in the log.
- `update_project` resolved slug uniqueness against the *caller's* namespace; slugs are unique per
  owner, so a contributor renaming a shared matter must be checked against the owner's.

## Config

`LQ_AI_MATTER_DEFAULT_SHARE_SCOPE` defaults to `personal` — fail-restrictive per ADR 0016 P4. A new
matter reaches only its owner until someone is added. Deployments where ambient visibility is the
working assumption set `org`.

## Docs

- `docs/security/matter-access-control.md` — operator-facing, including an explicit statement that
  this is application-layer authorization and **not** a cryptographic boundary, plus known
  limitations.
- ADR 0024 — the decision record. It says up front that it was written after the implementation
  rather than before; I'd rather record that honestly than back-date it.
- `docs/db-schema.md` — the table, constraints, and backfill.

## Not in this change

`audit_log.project_id` (so per-matter audit is still not producible — a real gap, worth its own
PR), optimistic concurrency on `context_md`, matter ownership transfer, team-scope grants, and
making file/KB visibility follow the matter.

## Testing

`pytest tests/` — 2709 passed. Two failures remain and **both reproduce on unmodified `main`**; I
ran a clean `origin/main` worktree to confirm rather than assume:

- `test_sighup_reload_in_child_process` segfaults the whole pytest process in a full run (passes
  alone) — a redundant `env=` takes CPython off `posix_spawn`.
- `test_receipt_assembles_phase_transitions_and_tool_calls` is flaky because
  `build_receipt` orders by `audit_log.timestamp`, which is *transaction* time, so same-transaction
  events are indistinguishable.

Happy to file those separately if useful.

New: `tests/test_matter_authz.py` (the access matrix, including the DB-level constraints and the
absent-capability tests), `tests/test_matter_members.py`, `tests/test_users_directory.py`.

---

### DIFF
--- a/api/alembic/versions/0067_matter_membership.py
+++ b/api/alembic/versions/0067_matter_membership.py
@@ -1,0 +1,163 @@
+"""project_members + projects.share_scope — matter-scoped collaboration
+
+Revision ID: 0067
+Revises: 0066
+Create Date: 2026-08-20
+
+PRD §3.11 lists ``share_scope`` in the Project data model and names
+``POST /api/v1/projects/{id}/share``; §3.11's M1 status records
+share-with-group as deferred. ADR 0020 D3 pins "matter = project — there
+is no new ``Matter`` model". This migration lands the deferred half:
+a membership table over ``projects`` and the ``share_scope`` column the
+PRD already named.
+
+**project_members** mirrors ``team_members`` (migration 0014) verbatim —
+composite PK, the same CASCADE/RESTRICT split, and the same
+``added_by_user_id`` forensic column so "who let X onto matter Y?"
+survives in audit-log queries that index by actor.
+
+The ``role`` enum extends the team model with one value: ``blocked``.
+A blocked row is a **negative** grant — an ethical screen / conflict
+wall — and the resolver evaluates it before every allow, so it overrides
+firm-wide scope, explicit membership, and operator-admin alike. Storing
+denial as a role value rather than a separate table keeps "who can see
+this matter, and who explicitly cannot" answerable from one indexed
+query.
+
+**share_scope** is the ambient grant:
+
+* ``personal`` — the owner and explicit members only (the pre-migration
+  behaviour, and the default upstream).
+* ``members`` — same as personal today; reserved so a deployment can
+  distinguish "deliberately restricted" from "never shared" in the UI
+  without a second column.
+* ``org`` — every non-blocked user in the deployment gets **read**.
+  Contributing still requires an explicit membership row, so the roster
+  stays a truthful answer to "who worked this matter" — which is the
+  question that matters for privilege and conflicts.
+
+Sandbox matters (``is_sandbox``) are per-user scratch space and are
+constrained to ``personal`` at the DB layer; sharing one would leak a
+colleague's try-it space into the matter list.
+
+**Backfill:** one ``lead`` row per existing project for its owner, so the
+owner path and the membership path are structurally identical from row
+one and the resolver returns exactly the pre-migration answer for every
+existing row. ``projects.owner_id`` is ``ON DELETE RESTRICT`` and
+non-nullable, so every project has a live owner to backfill from.
+
+Reversible: ``downgrade()`` drops the constraint, the column, and the
+table. Membership rows are lost, which is correct — without the table
+there is nothing to express them.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic
+revision: str = "0067"
+down_revision: str | None = "0066"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    # --- project_members ---------------------------------------------------
+    op.create_table(
+        "project_members",
+        sa.Column(
+            "project_id",
+            postgresql.UUID(as_uuid=True),
+            sa.ForeignKey(
+                "projects.id",
+                ondelete="CASCADE",
+                name="fk_project_members_project",
+            ),
+            nullable=False,
+        ),
+        sa.Column(
+            "user_id",
+            postgresql.UUID(as_uuid=True),
+            sa.ForeignKey(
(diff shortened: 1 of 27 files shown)


houfu:
Review update at `b49e0b5879a6618586dcbcedaf7b824620a0a401`: I think this still needs changes before merge. A few limitations can reasonably become follow-ups, but the items below contradict the core access-control and deletion guarantees of this PR.

1. **The current API failure is not either flaky failure described in the PR body.** `test_openapi_paths_match_sketch` expects `/api/v1/admin/users/{user_id}/reset-password`, but this PR does not provide that route. The run finished with 2,692 passed, 1 skipped, and this one failure. Please remove the unrelated expectation or make the dependency explicit and rebase once it exists.

2. **A screen does not revoke access to chats the screened user owns.** `_load_readable_chat` returns immediately when `row.owner_id == user.id`, before resolving matter access. The write path also remains owner-only and does not re-check the matter. A contributor can therefore create a chat on a shared matter, later be changed to `blocked`, and retain access to that matter-linked chat. The existing regression covers a screened user reading somebody else's chat, so it misses this case. Because the PR defines `blocked` as absolute, I consider this a merge blocker.

3. **A sandbox can still be shared through an explicit membership.** The database constraint only prevents `share_scope != 'personal'`; `add_member` accepts another user on a sandbox and `resolve_access` honors that membership. This conflicts with the documented statement that sandboxes are per-user and cannot be shared. Either reject roster mutations on sandboxes or revise the invariant and threat model. Given the current narrative, this is a blocker.

4. **A failed user deletion can remove object-store bytes permanently.** `_hard_delete_user` deletes export and file bytes before discovering shared matters that keep `projects.owner_id` and the user row under `RESTRICT`. The database transaction can roll back, but object storage cannot, leaving retained database rows whose bytes are gone. The shared-matter preflight needs to happen before irreversible storage deletion (and `added_by_user_id` restrictions need the same treatment).

I would be comfortable tracking ownership transfer, roster UI, and broader handling of already-running autonomous sessions as explicit follow-ups. I would not leave the screen bypass, sandbox contradiction, or deletion ordering unfixed, because they undermine the guarantees this PR is introducing.

Review basis: static review of this exact head plus the recorded CI output; I did not execute the contributor branch.

PR #537, opened 2026-08-21
fix: release image ships an empty skills/community

sergiomaldo:
`api/Dockerfile.release` does `COPY skills /skills` so the published image is
self-contained. Half that corpus lives in the `skills/community` submodule, but the
`build-backend` checkout does not request submodules, so the directory is empty at
build time.

`COPY` of an empty directory succeeds, so nothing fails: the image builds, starts
healthy, and is simply missing those skills. It only surfaces when someone reaches for
one of them.

**Changes**

- `.github/workflows/release.yml`: request `submodules: recursive` on the
  `build-backend` checkout only. No other checkout is touched.
- `api/Dockerfile.release`: fail the build unless at least one
  `skills/community/skills/*/SKILL.md` manifest is present. The runtime loads community
  skills from that path (`community_skills_dir` in `api/app/config.py`), so an
  emptiness check would pass on a submodule root that carries only README and LICENSE.
  The corpus now lives in its own `skills` build stage that the final stage copies
  from verbatim. The guard is unchanged in effect, but the stage can be built with
  `--target skills` in seconds, without the dependency install (which pulls torch via
  docling).
- `scripts/release-image-check.sh`: regression coverage for the release-packaging
  path, runnable locally. It asserts that a recursive checkout bakes the community
  manifests (count matches the checkout), and that an empty submodule directory, a
  metadata-only submodule root, and a `skills/` tree with no manifest each stop the
  build at the guard (it checks for the guard's own message, not just a non-zero
  exit). `FULL_IMAGE=1` also builds the complete api release image.
- `.github/workflows/ci.yml`: new `release-image-checks` job runs the script on every
  PR with a recursive checkout. Under a minute cold.
- `CONTRIBUTING.md`: document the check next to the stack smoke section.

**Review follow-up**

- Guard strengthened to require an actual `SKILL.md` manifest (inline comment).
- Regression coverage added for the three requested cases, plus a fourth (a slug
  directory without a manifest).
- Commit amended with `--signoff`.

**Verification (locally, on arm64)**

- `./scripts/release-image-check.sh`: all four cases pass in about 50 seconds
  including the base image pull. The `skills` stage carries 44 community and 15 firm
  manifests, matching the checkout.
- `FULL_IMAGE=1 ./scripts/release-image-check.sh`: the complete api release image
  carries the same 44 community manifests.

The guard only affects `api/Dockerfile.release`; `gateway/Dockerfile.release` does not
copy the corpus.

https://claude.ai/code/session_01Lx8YLABGnbDqhznyouzW4s

### DIFF
--- a/api/Dockerfile.release
+++ b/api/Dockerfile.release
@@ -4,6 +4,32 @@
 # bake the repo-root skills corpus (bind-mounted in dev) for the self-contained
 # launcher images; built with repo-root context by .github/workflows/release.yml.
 
+# Skills-corpus stage. Stands alone (no dependency install) so the guard below
+# can be exercised in seconds with `docker build --target skills`, which is
+# what scripts/release-image-check.sh and the CI job built on it do. The final
+# stage copies /skills from here verbatim, so what this stage guarantees, the
+# published image carries.
+FROM python:3.12-slim AS skills
+
+# Bake the skills corpus that the dev stack bind-mounts (./skills:/skills:ro),
+# so the published image is self-contained for the zero-checkout macOS launcher.
+COPY skills /skills
+# COPY of an empty directory succeeds silently, so a build from a checkout
+# without submodules produces an image that starts healthy and is simply
+# missing half the corpus. Fail the build instead.
+#
+# "Non-empty" is not enough: the runtime loads community skills from
+# /skills/community/skills/<slug>/SKILL.md (api/app/config.py,
+# community_skills_dir), so a checkout carrying only the submodule's README
+# and LICENSE would pass an emptiness check with zero loadable skills. Assert
+# that at least one manifest exists. If the glob matches nothing, $1 is the
+# literal pattern and `test -f` fails.
+RUN set -- /skills/community/skills/*/SKILL.md; test -f "$1" || { \
+      echo "ERROR: no community skill manifests found under /skills/community/skills/*/SKILL.md - the skills/community submodule was not checked out." >&2; \
+      echo "Run 'git submodule update --init --recursive' before building this image." >&2; \
+      exit 1; \
+    }
+
 FROM python:3.12-slim AS base
 
 # ADR 0023 — dependencies install from the committed uv.lock so the shipped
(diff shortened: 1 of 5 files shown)


api/Dockerfile.release
```diff
@@ -42,6 +42,14 @@ RUN chmod +x /usr/local/bin/lq-ai-entrypoint
 # Bake the skills corpus that the dev stack bind-mounts (./skills:/skills:ro),
 # so the published image is self-contained for the zero-checkout macOS launcher.
 COPY skills /skills
+# COPY of an empty directory succeeds silently, so a build from a checkout
+# without submodules produces an image that starts healthy and is simply
+# missing half the corpus. Fail the build instead.
+RUN test -n "$(ls -A /skills/community 2>/dev/null)" || { \
+      echo "ERROR: /skills/community is empty - the skills/community submodule was not checked out." >&2; \
```
houfu: The Docker guard currently checks only whether /skills/community is non-empty. The runtime loads skills from /skills/community/skills/<slug>/SKILL.md, so repository metadata such as a README could make the check pass even when no community skills are loadable. Please validate that at least one actual skills/*/SKILL.md file exists.

houfu (changes requested):
Thanks for spotting this. Just a few minor changes to make this PR solid.

1.  Please add regression coverage for the release-packaging path. The current API, Gateway, and Web checks do not build api/Dockerfile.release. Ideally, the test should verify that:
* a recursive checkout produces an image containing community SKILL.md files;
* missing community content fails the build; and
* a non-empty submodule root without any actual skill manifests also fails.
2. The commit is missing the required DCO Signed-off-by trailer. Please amend the commit with --signoff and update the branch.

api/Dockerfile.release
sergiomaldo: Done. The guard now asserts that at least one `/skills/community/skills/*/SKILL.md` exists (a glob that matches nothing leaves the literal pattern in `$1`, and `test -f` fails). The corpus moved into its own `skills` stage so the guard can be built with `--target skills` in seconds, and `scripts/release-image-check.sh` covers the metadata-only case you describe, plus a slug directory without a manifest.

houfu (approved):
Thanks for the thorough update. The release-image regression coverage now exercises the successful recursive checkout and the relevant failure cases, and the commit includes the required DCO sign-off.

I’ve re-reviewed the workflow, Dockerfile, and test-script changes. The implementation fails safely when community skills are unavailable, and all checks are green. Approved — ready to merge.

